Issue #2, opened 2015-02-18
Encode scopes and variables in source map


fitzgen:
Source maps should be able to encode the original source language's environment, scopes, and variables. They should be encoded in such a way that if a debugger is paused at a given generated JS location, it can restore and display the original source language's variables and parent scopes that are in scope at the paused location.

fitzgen:
Here are some cases that I think the source map's environment rematerialization
should support:

* When there are scopes in JS that do not correspond to any scopes in the
  original source language. For example, the compiler emitted an
  immediately-invoked-function-expression as an implementation detail that
  doesn't reflect any nested function or scope in the original source language.

* When there are scopes in the original source language that do not correspond
  to any scope in the generated JS code. For example, one could imagine an
  ES6 to ES3 compiler transforming this ES6 code:

        {
          let x = 1;
          console.log(x);

          {
            let x = 2;
            console.log(x);
          }

          console.log(x);
        }

  Into this ES3 code:

        {
          var x1 = 1;
          console.log(x1);
          var x2 = 2;
          console.log(x2);
          console.log(x1);
        }

  Note the nested block scope in the ES6 source that does not exist in the ES3
  source. We should be able to recreate this scope.

* We should support simple variable renaming.

  For example, a Scheme-to-JS compiler might emit variables with `!` replaced as
  `_bang`: `set-cdr!` becomes `set_cdr_bang`.

  Another example: an ES6 to ES3 compiler might bind the `this` outside of an
  arrow function to a variable and close over it:

        var myArrow = () => this.x;

                |
                V

        var _this = this;
        var myArrow = function () { return _this.x; };

  The rematerialized scope inside the `myArrow` function should have a `this`
  binding that points to the `_this` variable.

* We should support hiding bindings in the generated JS code that do not
  correspond to any bindings in the original source language. These might be
  gensyms, or temporary variables, or any implementation detail of the
  compiler's emitted JS code.

* We should support rematerializing bindings in the original source language
  that do not have any corresponding bindings in the generated JS code.

  First example (Python to JS):

        result = [x + 1 for x in list]

                |
                V

        var result = [];
        for (var i = 0; i < list.length; i++) {
          // Note: no `x` binding in this generated JS code.
          result.push(list[i] + 1);
        }

  Second example (C++ ish to JS, psuedo-code for brevity):

        class Point {
          int x;
          int y;

          // `Point lhs += Point rhs` == `lhs.x += rhs.x; lhs.y += rhs.y;`
        }

[...]

PR #4, opened 2015-07-22
 Proposal for encoding source-level environment information

fitzgen:
Proposal for #2.
### [Rendered Proposal](https://github.com/fitzgen/source-map-rfc/blob/scopes-and-bindings/proposals/env.md)

I've tried to make this precise and detailed, perhaps at the cost of some brevity, but I was also not 100% sure how detailed and specific to be at times. I am very open to suggested changes, particularly anything that is not clear to potential implementers.

Looking forward to feedback.

### DIFF
--- a/proposals/env.md
+++ b/proposals/env.md
@@ -1,0 +1,591 @@
+# Proposal for Encoding Source-Level Environment Information Within Source Maps
+
+Author: Nick Fitzgerald <fitzgen@mozilla.com>, Mozilla
+
+Date: July, 2015
+
+This work is licensed under a
+[Creative Commons Attribution-ShareAlike 4.0 International License](https://creativecommons.org/licenses/by-sa/4.0/).
+
+## Abstract
+
+This document describes a proposed extension to the
+[source map format](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1#)
+for encoding source-level environment debugging information, such as scopes and
+bindings.
+
+## Introduction
+
+When boiled down to their essence, stepping debuggers enable users to do two
+things:
+
+1. Set breakpoints in a program's source and incrementally step through its
+   execution.
+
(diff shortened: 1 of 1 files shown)


fitzgen:
cc @andysterland @concavelenz @swannodette

yurydelendik:
Here is a couple of thoughts related to the information expressed in encoded scopes. The first one is that scopes will be better expressed in "source location" locations. The good example of the transform, where it will be useful, is a function inlining -- the function scope is not really belong to the chain in the generated scopes where the body is inlined. (Bound locals/params need to rely on generated locations, see the thought below) It might be useful to have mapping of the generated scopes mapped to original locations to avoid guessing, when debugger is trying to resolve binding values.

The second thought is to replace BindingValue to something more like an expression, e.g. encoding folded constant, locals in an unrolled loop, or an eliminated induction variable. Notice that a binding may change its value in the same scope within generated source -- it can complicate value resolution though. Expression-like values is useful to build more "structural" value e.g. for asm.js/wasm generated from C/C++.

proposals/env.md
```diff
@@ -0,0 +1,591 @@
...
+(again, similar to DWARF). This allows for the definition of an abbreviation
+that includes a record type and the set of properties. Then, when serializing a
+record whose type and property set already has an abbreviation definition, the
+record's type and properties can be omitted, emitting only the property
+values. Additionally, a property's value is encoded relative to the last value
+emitted for that property. This technique is used in encoding segments in the
+`"mappings"` property of the source map format, and has proved valuable in
+reducing file size.
+
+The `"env"` property can be parsed with the following BNF-like grammar, where
```
stefanpenner: Concatenation of source files, into final output files is extremely common. One proposed aspect of sourceMap, are sections. Those haven't been adopted, but if they would. It would simplified by the concatenation tooling implementations (and performance) and I believe provide similar benefit to debugging tools utilizing source-maps. Given then, should `env` also have `envSections` or something similar?

PR #59, opened 2023-10-27
Add encoding to scopes proposal

hbenl:
- copied from https://gist.github.com/sokra/97a53a869b9a421accadbc9681cb26f3
- removed the `callsite`'s end location
- removed the `expressions` field
- added some clarifications
- fixed the example
- removed the outdated JSON example


littledan (approved):
Landing because no concerns were raised over a week, so it's easier for others to review.

vobjhss:
I'm located in Atlanta, Ga. Only Medicare calls come to my device. I'm not receiving any messages. Why is my napping area blocked

PR #60, opened 2023-11-08
Add range mappings proposal

sokra:
Rendered view: https://github.com/tc39/source-map-rfc/blob/sokra/range-mapping/proposals/range-mappings.md

### DIFF
--- a/proposals/range-mappings.md
+++ b/proposals/range-mappings.md
@@ -1,0 +1,91 @@
+# Range Mappings
+
+* Author: Tobias Koppers
+* Date: November, 2023
+
+## Motiviation
+
+Currently mappings map locations in the source code to other locations in the source code.
+This works well when trying to access a location that is defined in the SourceMap, but one looses percision when accessing locations that are not directly defined in the SourceMap.
+In these cases tools usually fallback to next lower column that is actually mapped in the SourceMap.
+So we are either loosing information or we need many mappings in the SourceMap to cover all possible locations.
+
+These information problem is especially problematic when applying a SourceMap to another SourceMap.
+Here we can only use locations that are specified in both SourceMaps. We have to be lucky that locations match up.
+
+## Proposal
+
+Add a boolean flag for each mapping to convert it into a "range mapping".
+For a range mapping, tools should assume that every char that follows the mapping (until the next mapping), is mapped to the specified original location plus the offset in the generated code.
+This means all chars in the generated code that is covered by the range mapping, are mapped char by char to the same range in the original code.
+(Usually this only makes sense when generated and original are identical for that range)
+
+### Example
+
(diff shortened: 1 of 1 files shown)


petebacondarwin (commented):
This is a great idea and would really help with higher fidelity merging of stacked source-maps.

jkrems:
I'm currently writing some code where I'm missing exactly this feature (efficient source maps for low transformation situations).

littledan:
> (Usually this only makes sense when generated and original are identical for that range)

Would we consider “invalid” a source map where this property is not met?

szuend:
The main motivation for the proposal is the composition of multiple source maps. Nonetheless I checked where DevTools would benefit from range mappings. Range mappings don't unlock any major features but would make smaller ones more accurate. DevTools often uses the closest mapped point if we try to resolve an unmapped point. In particular the following could get better:

- Setting inline breakpoints.
- Exact call location for stack traces (in cases where the call point is not mapped)
- When hovering over an expression while paused without the "Consistent Source Map Variables" experiment enabled.

jridgewell:
I think we could optimize the encoding format a bit more by using lengths instead of bit-per-mapping. This would be particularly beneficial for maps with largely unchanged outputs that contain many mappings.

Ie, consider internal state `range` (defaulting to `false`) and `rangeIndex` (defaulting to `0`) for every source map. Every mapping encountered is considered normal while `range` is `false`. In order to flip `range`, emit a VLQ integer that increments the current `rangeIndex`. The mapping at that index and every mapping afterwards is considered a range mapping. In order to flip `range` back, emit another VLQ integer. Every mapping after that index is a normal mapping.

Our example `rangeMappings` would just be `aCMC` instead of `AAB;;g` (13th index turns on, 14th index turns off, 20th index turns on, 21st turns off). In order to turn an entire source map into a range map would be just `A`.

sokra:
@jridgewell Possible. I would say it's a trade-off. It might be less efficient when range mappings and normal mappings alternate a lot. It would expect that for e. g. minimizer produced source maps.

My thinking behind the encoding I proposed was:

* Separated by lines with `;` allows to easily scan by line or split parsing by line. It's also easy to concatenate source maps as `rangeMappings` could just be concatenated.
* Bitwise encoding: It's very cheap and deterministic. Best case (single line): 1.33bits/0.17B per mapping, worse case (1 mapping per generated line): 16bits/2B per mapping. It mostly depend on the number of generated lines, which is likely to be very low in production scenarios where code in minified.
* It's very cheap to decode, an optimized implementation might not decode it at all and only maintain a map that allows to find the start of a line cheaply.
* That's possible because format allows very cheap random access to the bits without fully decoding

jkup (approved):

PR #66, opened 2024-02-08
scopes: Rename GeneratedScope to GeneratedRange

szuend:
Small PR to rename "GeneratedScope" to "GeneratedRange" to make it more clear that it does not need to correspond to a true lexical JS scope. Alternatively we can just call this out explicitly but leave the "GeneratedScope" name. Let me know what you prefer.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -113,13 +113,13 @@ The following information describes a scope in the source map:
   - a javascript expression that can be evaluated by the debugger in the corresponding generated scope to get the binding's value (if such an expression is available)
 
 The following code snippet specifies the scope information **conceptually** in TypeScript notation. See the [Encoding](#encoding) section
-on how this information is actually VLQ encoded.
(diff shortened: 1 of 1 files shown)


proposals/scopes.md
```diff
@@ -113,13 +113,13 @@ The following information describes a scope in the source map:
...
 The following code snippet specifies the scope information **conceptually** in TypeScript notation. See the [Encoding](#encoding) section
-on how this information is actually VLQ encoded.
+on how this information is actually VLQ encoded. We chose the name `GeneratedRange` instead of `GeneratedScope` to make it explicit that a `GeneratedRange` does not necessarily correspond to a lexical ECMAScript scope (e.g. in the case of an inlined function body).
 
 ```ts
 interface SourceMap {
   // ...
   originalScopes?: OriginalScope[];
-  generatedScopes?: GeneratedScope;
+  generatedRanges?: GeneratedRange;
```
jkup, line 122: Should this be `generatedRange` singular if it only takes a single range?

hbenl: `GeneratedRange` contains child ranges.

hbenl (approved):

proposals/scopes.md
szuend: I can see how both names fit. I'll land as is but we can bike-shed the names in the future :)

PR #68, opened 2024-02-27
Remove hasBindings and bindings count from scopes encoding

hbenl:
We don't need to include the number of variables/bindings in the encoded scope information: variables/bindings are added at the end of each "Start Scope" item and so the sourcemap consumer can simply decode variables/bindings until it reaches the end of that item.
So we can make the encoded scopes a little smaller by removing the number of variables/bindings. (The only possible downside is that we lose the potential to extend the current format in a backward compatible way by adding information to the end of a "Start Scope" item)

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -203,14 +203,12 @@ Note: Each DATA represents one VLQ number.
   * Note: binary flags that specify if a field is used for this scope.
   * Note: Unknown flags would skip the whole scope.
   * 0x1 has name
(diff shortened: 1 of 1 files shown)


szuend (approved):

jridgewell:
Another loss here is the ability to reallocate an array of the appropriate size. Individually pushing variables into and array can leave it much longer capacity than necessary: https://github.com/Rich-Harris/sourcemap-codec/pull/80

Issue #35, opened 2023-04-13
Exit criteria for a hardened specification


littledan:
For our goals (#22), I see three independent work streams which do not need to block each other: Hardening the specification can take place in parallel with adding features to improve naming and association of source files, though the latter will benefit from the solid framework of the former.

In this issue, let's discuss what sorts of things we'd want to see from a specification before we can tell the community, "look over here, this is now the 'real' source map specification". I believe this depends on "hardening" and not on addition of features (but we can still work on both at the same time).

What I'd consider hard requirements:
- Unambiguous specification text which clarifies the questions that have been raised so far in the repository, e.g., what do columns mean, how are the URLs interpreted, etc.
- Some kind of shared tests for both source map generators and consumers
- A license ensuring that it's safe to implement and use the source maps specification
- The specification matches implementations generally, modulo bugs, and we're confident that it will make sense to report any deviations from the specification to tools authors as bugs that they should fix.

My preferences that we also get the following in place, but I don't think that these necessarily block us telling the community that they should refer to our specification:
- An inclusive process for making changes and additions going forward, on an ongoing basis, as needed
- A home in a standards body (probably TC39, or else WHATWG or W3C) which will provide backing over time for both the IPR license and the inclusive evolution process
    - I don't think going through all of the motions of full formal standardization should be a precondition to advising the community to reference our definition of source maps.

PR #71, opened 2024-03-04
Link to TC39 spec

jridgewell:


nicolo-ribaudo (approved):

nicolo-ribaudo:
Thanks!

littledan:
Huh? I thought we were going to keep linking to the previous spec until we met certain exit criteria #35 . We can link to the under-development spec, but we shouldn't identify it as v3 at this point.

jridgewell:
Before we publish our own specification, sure it'd make sense to hit those exit criteria. But I think it's completely reasonable for our RFC repo to directly link to our spec repo.

nicolo-ribaudo:
Isn't this like linking to tc39.es/ecma262 rather than to a published version? i.e. the draft source maps spec is already an improvement over the Google doc, but it's still not published and marked as draft (until we meet the exit criteria)

littledan:
Yeah, I wanted us to get to those exit criteria before publishing something equivalent to tc39.es/ecma262 (which is viewed as an authoritative reference, as it should be, even though it doesn't have official status) and then we can continue updating this draft as a "living standard".

What I would prefer is if we did link to this draft, but identified it as "not 100% ready yet", and also linked to the previous v3 spec, and identified it as "historical"--this patch just identifies the current draft as v3, which seems confusing.

Anyway, maybe we should consider ourselves to have met whatever minimum exit-to-promoted-draft milestone; I would have just preferred explicit discussion/acknowledgement of this.

nicolo-ribaudo:
> What I would prefer is if we did link to this draft, but identified it as "not 100% ready yet", and also linked to the previous v3 spec, and identified it as "historical"

In the first few lines of the draft document it says that it's a strawman and it links to the historical version -- do you suggest to also keep both links in the readme?

> Anyway, maybe we should consider ourselves to have met whatever minimum exit-to-promoted-draft milestone; I would have just preferred explicit discussion/acknowledgement of this.

Yes I agree we should discuss together whether we meet some minimum requirement yet, but I disagree that linking to the draft we're working on in this repo is marking it as "ready" in any way.

Anyway, let's revert this and discuss it at the next meeting?

PR #74, opened 2024-03-06
Revert "Link to TC39 spec"

nicolo-ribaudo:
Reverts tc39/source-map-rfc#71, see discussion in that PR. I'll open a revert of the revert so we can discuss at the next meeting.


jkup (approved):

jkup:
We moved this file to README.rfcs.md during the repo combination. I'll just make a new PR that does this revert but as a new commit against the right file if that's ok!

PR #82, opened 2024-04-19
 Fix "This version" URL pointing to former source-map org

Krinkle:
The old URL serves 404 Error now.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -4,7 +4,7 @@ H1: Source Map
 Shortname: source-map
 Level: 1
 Status: STAGE0
-URL: https://source-map.github.io/source-map-spec/
+URL: https://tc39.github.io/source-map-spec/
 Editor: Armin Ronacher, Sentry
 Former Editor: Victor Porof, Google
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -4,7 +4,7 @@ H1: Source Map
 Shortname: source-map
 Level: 1
 Status: STAGE0
-URL: https://source-map.github.io/source-map-spec/
+URL: https://tc39.github.io/source-map-spec/
```
jkrems, line 7: I think I would go all the way and resolve the redirects to https://tc39.es/source-map/.

jkrems (changes requested):
Thanks for finding this!

source-map.bs
Krinkle: Okay.

jkrems (approved):

jkup (approved):

jkup:
Thank you! Sorry this sat for so long, I was struggling to understand the IPR form check. Appreciate the PR!

Issue #70, opened 2024-02-27
Scopes: how to encode unavailable bindings


hbenl:
The scopes proposal adds information about scopes and their variables and bindings to the sourcemap. But some bindings will be unavailable, i.e. for some original variables there will be no expression that can be evaluated to get the variable's current value. We could encode this by using `-1` as the index in the `names` array for these bindings but this clashes with our use of *relative* indices into the `names` array. We use relative indices in the hopes of reducing the encoded scope's size because the encoding of relative indices will hopefully be smaller on average than that of absolute indices. But if some of the indices will be `-1` then using relative indices will probably not reduce the size of the encoded scope.

szuend:
To summarize the possible solutions that were discussed on the call [2024-02-29](https://github.com/tc39/tg4/blob/main/meetings/2024/2024-02-29.md):

* Use an additional bitfield to indicate which of the live ranges indicate an "unavailable" range and the expression index must be ignored.
* Introduce one more character on top of `"` and `,` to signal `undefined` or `null`.

rakudrama:
There is a VLQ encoding that is short and useless as a signed increment: `B` encodes -2<sup>31</sup>.
This could be repurposed to represent `null` with a zero increment.

There would be a new operation *_RelativeNameOrNull(vlq, accumulator, names)_*.
Same as *_RelativeName_* except for one new line between step 1 and step 2:

>  1.5 if relativeIndex = -2<sup>31</sup> return null

if we just updated _RelativeName_ with this line, is there a source map file that would mean something different?

szuend:
We use unsigned VLQs for binding expressions so `B` is not available: https://tc39.es/ecma426/branch/proposal-scopes/#sec-BindingExpression. We'd have to sacrifice 1-bit in each VLQ to encode a single value, which we considerd not worth it.

PR #83, opened 2024-04-25
Scopes: Make the offsets into the names array absolute

hbenl:
At the last meeting we decided to make the offsets into the names array absolute for now until we have decided how to deal with #70 .
@jridgewell I also wanted to add you as a reviewer but for some reason Github won't let me.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -206,11 +206,10 @@ Note: Each DATA represents one VLQ number.
   * 0x1 has name
 * name: (only exists if `has name` flag is set)
   * DATA offset into `names` field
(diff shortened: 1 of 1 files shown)


proposals/scopes.md
```diff
@@ -206,11 +206,10 @@ Note: Each DATA represents one VLQ number.
   * 0x1 has name
 * name: (only exists if `has name` flag is set)
   * DATA offset into `names` field
```
jridgewell, line 208: Should we clarify that these are absolute offsets?

jridgewell (approved):

szuend (approved):

PR #87, opened 2024-05-31
Mark names array as optional

jkup:


szuend (approved):

jridgewell (approved):
My code treats it as optional already.

PR #90, opened 2024-06-11
Fix link to WASM Custom Section

jkup:
@JSMonk it looks like the w3 link we have is no longer working, is this the one we should be using?

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -65,7 +65,7 @@ spec:url; type:dfn; for:/; text:url
     "title": "Source Map Revision 2 Proposal"
   },
   "WasmCustomSection": {
-    "href": "https://www.w3.org/TR/wasm-core-2/binary/modules.html#custom-section",
+    "href": "https://www.w3.org/TR/wasm-core-2/#custom-section%E2%91%A0",
     "publisher": "W3C",
     "status": "Living Standard",
     "title": "WebAssembly custom section"
--- a/source-map.bs
+++ b/source-map.bs
@@ -65,13 +65,13 @@ spec:url; type:dfn; for:/; text:url
     "title": "Source Map Revision 2 Proposal"
   },
   "WasmCustomSection": {
-    "href": "https://www.w3.org/TR/wasm-core-2/#custom-section%E2%91%A0",
+    "href": "https://webassembly.github.io/spec/core/binary/modules.html#custom-section",
     "publisher": "W3C",
     "status": "Living Standard",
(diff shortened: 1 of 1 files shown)


JSMonk:
Yep, seems like. Thank you for the update

jkup:
Actually it looks like maybe the binary format was moved to https://webassembly.github.io/spec/core/binary/modules.html#custom-section

source-map.bs
```diff
@@ -65,7 +65,7 @@ spec:url; type:dfn; for:/; text:url
     "title": "Source Map Revision 2 Proposal"
   },
   "WasmCustomSection": {
-    "href": "https://www.w3.org/TR/wasm-core-2/binary/modules.html#custom-section",
+    "href": "https://www.w3.org/TR/wasm-core-2/#custom-section%E2%91%A0",
```
JSMonk, line 68: Shouldn't it be just `https://www.w3.org/TR/wasm-core-2/#custom-section`?

jkup: Should we link to the https://www.w3.org/TR/wasm-core-2/#custom-section or the https://webassembly.github.io/spec/core/binary/modules.html#custom-section

JSMonk: There is no difference in content. I usually check the second variant of the specification (because of the less contrast colors). The only think came into my mind, that the `w3.org` version looks more official and from the official specification it would be better to mention other official specification. 
But what do you think fit better?

jkup: That makes sense to me, let's go with the w3.org one because it does look nice and official!

JSMonk (approved):

source-map.bs
```diff
@@ -65,13 +65,13 @@ spec:url; type:dfn; for:/; text:url
...
   "WasmCustomSection": {
-    "href": "https://www.w3.org/TR/wasm-core-2/binary/modules.html#custom-section",
+    "href": "https://www.w3.org/TR/wasm-core-2/#custom-section",
     "publisher": "W3C",
     "status": "Living Standard",
     "title": "WebAssembly custom section"
   },
   "WasmNamesBinaryFormat": {
-    "href": "https://www.w3.org/TR/wasm-core-2/binary/values.html#names",
+    "href": "https://www.w3.org/TR/wasm-core-2/#names%E2%91%A2",
```
JSMonk, line 74: The same here, shouldn't it be just `https://www.w3.org/TR/wasm-core-2/#names`?

nicolo-ribaudo (approved):

PR #91, opened 2024-06-11
Revert spec change

jkup:
We changed the linked spec from the google doc to the current draft but received feedback we should revert for now. Between the change and the revert we combine the repositories so the revert got a bit messy. This reverts to the original link to the google doc as the spec.


jkup:
I don't feel strongly about this. I've heard from one person that it's important to only link to the formalized spec until our draft gets approved. On the other hand I think our draft is far more helpful to developers than the google doc. Happy to go either way here, I suppose I have a personal preference for leaving the link to our draft -- as long as that isn't directly against and TC39 rules.

PR #94, opened 2024-06-13
Define how to extract the `sourceMappingURL` comment

nicolo-ribaudo:
This was originally opened at https://github.com/tc39/source-map-spec/pull/30

I am currently being hand-wavy about CSS, only saying "it should be similar to JS". I can propose the adjusted algorithm in a followup, but given that technically source maps are not language-specific we might also just say "other text languages should be like JS, adapted to their own comments syntax".

---

This patch explicitly defines how to extract such comments from JavaScript, CSS and WebAssembly sources.

It defines multiple ways to do so: either by actually parsing the code, or by just going through all the lines of the program looking for what "looks like" a comment. This is so that different implementations can choose what's best for them, depending on whether they are already parsing the code or not.

To ensure consist behavior accross implementations that choose different strategies, the specification enforces additional requirements on tools that append a `sourceMappingURL` comment to the generated code: the comment must be placed in such a way that all extraction methods yield the same result. This is not an unresonable burden, since if the progeram is syntactically valid, simply adding the comment at the end of the file only potentially followed by other tool-injected comments is enough. This requirement is lifted if the input code given to the tool is already "maliciously crafted", since we would otherwise require tool to go rewrite that code (for example, splitting strings that contain something that looks like a comment).

It has the following properties:
- It iterates line by line. Implementations can thus optimize it by going through each line _in reverse order_, and then scanning through its characters from the beginning to the end (which is what a regexp would do).
- It expects multi-line comments to actually be in a single line.
- It returns the last `sourceMappingURL` comment (or well, comment-like) found in the source.
- It only considers comments after the last piece of code (i.e. it discards any comment found so far every time it sees some non-comment non-whitespace characters).
- It has no requirements about what is _before_ a comment. Adding the comment at the end of the file without first ensuring that there is a newline before it is valid.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -562,7 +562,8 @@ a [=byte sequence=] |bytes|, run the following steps:
 1. Return null.
 
 Since WebAssembly is not a textual format and it does not support comments, it supports a single unambiguous extraction method.
-The URL is encoded using [[WasmNamesBinaryFormat]], and it's placed as the content of the [=custom section=].
+The URL is encoded using [[WasmNamesBinaryFormat]], and it's placed as the content of the [=custom section=]. It is invalid for
+tools that generate WebAssembly code to generate two [=custom section=] with the "sourceMappingURL" name.
 
 Linking eval'd code to named generated code
 -------------------------------------------
--- a/source-map.bs
+++ b/source-map.bs
@@ -563,7 +563,7 @@ a [=byte sequence=] |bytes|, run the following steps:
 
 Since WebAssembly is not a textual format and it does not support comments, it supports a single unambiguous extraction method.
 The URL is encoded using [[WasmNamesBinaryFormat]], and it's placed as the content of the [=custom section=]. It is invalid for
-tools that generate WebAssembly code to generate two [=custom section=] with the "sourceMappingURL" name.
+tools that generate WebAssembly code to generate two or more [=custom section|custom sections=] with the "sourceMappingURL" name.
 
 Linking eval'd code to named generated code
 -------------------------------------------
--- a/source-map.bs
+++ b/source-map.bs
@@ -24,12 +24,29 @@ spec:html; type:element;
     text:title
     text:link
 
-spec:bikeshed-1; type:dfn; for:railroad; text:optional
-
-spec:fetch; type:dfn; for:/; text:request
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -394,6 +386,182 @@ When the [=Source Mapping URL=] is not absolute, then it is relative to the gene
...
+run the following steps:
+
+1. Let |lines| be the result of [=strictly split|strictly splitting=] |source| on [=line
+    terminator code points|ECMAScript line terminator code points=].
+1. Let |lastURL| be null.
+1. [=For each=] |line| in |lines|:
+    1. Let |position| be a [=position variable=] for |line|, initially pointing at the start of |line|.
+    1. [=While=] |position| doesn't point past the end of |line|:
+        1. [=Collect a sequence of code points=] that are [=white space code points|ECMAScript
+            white space code points=] from |line| given |position|.
```
nicolo-ribaudo, line 456: In the original PR there was this comment by @gibson042:

> Is it a problem that ECMAScript white space is subject to change over time as future Unicode editions change the set of code points in general category "Space_Separator"?

I think it's ok to expect implementations to evolve together with Unicode, but how do other folks feel?

jkup (approved):
This looks really good! I'm keen to get the changes in so we can continue hardening this part of the spec. I think we're definitely ok with updating the spec if JavaScript adds more spaces to their spec.

source-map.bs
```diff
@@ -394,6 +386,182 @@ When the [=Source Mapping URL=] is not absolute, then it is relative to the gene
...
+To <dfn export>extract a Source Map URL from JavaScript through parsing</dfn> a [=string=] |source|,
+run the following steps:
+
+1. Let |tokens| be the [=list=] of [=tokens=]
+    obtained by parsing |source| according to [[ECMA-262]].
+1. [=For each=] |token| in |tokens|, in reverse order:
+    1. If |token| is not a [=single-line comment=] or a [=multi-line comment=], return null.
+    1. Let |comment| be the content of |token|.
+    1. If [=match a Source Map URL in a comment|matching a Source Map URL in=]
+        |comment| returns a [=string=], return it.
```
takikawa, line 444: Is a fallback case needed here that just does `return null` if the loop terminates with no return?

jkup (approved):

Issue #80, opened 2024-03-28
 Make 32-bit limit on VLQs more precise


takikawa:
Currently the [terminology](https://tc39.es/source-map/#terminology) section of the spec mentions that the VLQ Base64 encoded are limited to 32-bit quantities. 

This definition should probably be made more precise, and describe what "limited" means. For example, it could specify that such VLQs are invalid and implementations may reject them (other possible semantics might include clamping values to 32-bits, but I'm not aware of browser implementations that do this).

Firefox seems to [consider such VLQs invalid](https://github.com/mozilla/source-map/blob/master/wasm-mappings/src/lib.rs#L592) while WebKit & Chrome don't appear to enforce the limit [1](https://github.com/WebKit/WebKit/blob/main/Source/WebInspectorUI/UserInterface/Models/SourceMap.js#L240), [2](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.ts;l=814).

There's also a question of if the sign bit of the VLQ should be included in the 32 bits. It seems natural to me to include it, as a reason to pick a 32 bit limit to begin with might be to ensure an implementation can use a 32-bit representation to store the whole VLQ. However, I think Firefox actually doesn't include the sign bit in the check linked above (if it checked against `i32::MAX` instead of `u32::MAX` it would IIUC).

PR #98, opened 2024-06-24
Add language clarifying the 32-bit limit for VLQs

takikawa:
This PR addresses issue #80. I'm adding it here to have for discussion when we go through all the spec issues in the hackfest.


jkup (approved):

jridgewell (approved):

lforst:
I was wondering, what should tools do in case the limit is exceeded? Should we define a behavior or is that up to the implementers? I guess it is different for generators than debuggers. Generators could crash or noop while debuggers could simply reject the sourcemap.

PR #99, opened 2024-06-24
Try fixing IPR check

nicolo-ribaudo:


nicolo-ribaudo:
The action is failing because it's still running with the version from `main`

PR #102, opened 2024-06-25
Change our process to group stack trace rewriters and debuggers

jkup:
We've been having a bit of discussion at the hackathon and I'd like to be more transparent with our thinking. We've been looking at the [scopes proposal](https://github.com/tc39/source-map/blob/main/proposals/scopes.md) and as it nears completion, it will be our first proposal to go through our [process](https://github.com/tc39/source-map/edit/main/PROCESS.md). A few things occurred to us:

1. We should really try to establish a closer relationship with more stack trace rewriters. @nicolo-ribaudo said he would reach out to some folks and see if they've had any chance to look at our proposal yet.
2. Our "constituencies" language of "generators", "debuggers" and "Error monitoring tools" is proving a bit difficult to work with. There will surely be proposals like "custom formatters" that will simply not affect error monitoring tools. So we should alter our Stage 4 process to account for this.
3. @bmeurer had a good idea that perhaps we can merge those two categories into just a "debuggers" category (dropping the word "interactive"). This would give us some flexibility that only the constituents affected by a proposal have to implement it.
4. This also lead to a deeper conversation about the # of constituents required for stage 4. I think this last week has really shown us the [breadth](https://github.com/jkup/source-map-users) of source map users in the ecosystem. I've left it at 2 and 2 for now but would be happy to change it.
5. @sokra also pointed out that it might be nice to have some sort of criteria around an end to end implementation of a new feature. Proving that, for example, we can generate scopes with Babel/gen-mapping and consume those exact scopes with Chrome/Firefox.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,8 +27,7 @@ Thanks for helping out in the [source map effort](https://ecma-international.org
 We've been thinking about _source map constituencies_ in the following terms. Please feel free to suggest better titles or categorization methods!
 
 1. Generators. Bundlers, transpilers, compilers. Tools that have access to the source code and emit (among other things) source maps.
-2. Debuggers. Browsers and [standalone debuggers](https://www.replay.io/). These are tools that have access to the source map and the generated code.
-3. Error monitoring tools. Tools like Sentry and New Relic. These tools have access to an error stack trace and a source map.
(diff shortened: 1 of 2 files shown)


littledan (approved):

jridgewell (approved):

lforst:
As someone from the "Stack trace validators" group this seems fine because we are likely covered by the debuggers group as a more strict validator already 👍

nicolo-ribaudo (approved):

PR #110, opened 2024-06-25
First pass at spec review

jkup:
@nicolo-ribaudo @takikawa and I took an initial pass at reading and making changes to the spec. Just trying to make it easier to read. I still want to go through the second half but these seemed like nice changes.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -4,8 +4,8 @@ H1: Source Map
 Shortname: source-map
 Level: 1
 Status: STAGE0
-URL: https://source-map.github.io/source-map-spec/
-Editor: Armin Ronacher, Sentry
+URL: https://tc39.es/source-map/
+Editor: Asumu Takikawa, Igalia 
 Former Editor: Victor Porof, Google
 Former Editor: John Lenz, Google
 Former Editor: Nick Fitzgerald, Mozilla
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (approved):

source-map.bs
```diff
@@ -149,22 +144,23 @@ Note: The values that can be represented by the VLQ Base64 encoded are limited t
...
+generated code.  The definition for columns in source maps can depend on
+the content type.
+
+  * For JavaScript and CSS based source maps are defined to be in
+    UTF-16 code units analogous to JavaScript string indexes.  That means that
+    "A" (`LATIN CAPITAL LETTER A`) measures 1 code unit, and "🔥" (`FIRE`) measures
+    2 code units.
+  * For WebAssembly, columns are defined as byte offsets from the beginning 
+    of the binary content (and there is only one group representing a line).  Source maps
+    for other content types might diverge from this.
```
jridgewell, line 157:
```suggestion
    of the binary content (and there is only one group representing a line).
  * Source maps for other content types might diverge from this.
```

This seems like a separate point.

source-map.bs
```diff
@@ -187,43 +183,31 @@ following structure:
...
-<dfn><code>mappings</code></dfn> a string with the encoded mapping data (see [[#mappings-structure]]).
-
-<dfn><code>ignoreList</code></dfn> an optional list of indices of files that
-should be considered third party code, such as framework code or bundler-generated code. This 
-allows developer tools to avoid code that developers likely don't want to see
-or step through, without requiring developers to configure this beforehand.
-It refers to the [=sources=] array and lists the indices of all the known third-party sources
-in the source map. Some browsers may also use the deprecated <code>x_google_ignoreList</code>
-field if <code>ignoreList</code> is not present.
+* <dfn><code>version</code></dfn> is the version field which must always be the number `3` as an integer. The source map may be rejected in case of a value different from `3`.
```
jridgewell, line 186: Are we using line wrapping?

jridgewell (approved):

Issue #113, opened 2024-07-02
Scopes: support for hiding new injected scopes from the stack trace


nicolo-ribaudo:
While we focused a lot on function inlining, sometimes Babel does the opposite. For example, given this input:
```js
for (let x of arr) {
  console.trace(x);
  run(() => x);
}
```
Babel generates this code:
```js
var _iterator = _createForOfIteratorHelper(arr),
  _step;
try {
  var _loop = function _loop() {
    var x = _step.value;
    console.trace(x);
    run(function () {
      return x;
    });
  };
  for (_iterator.s(); !(_step = _iterator.n()).done;) {
    _loop();
  }
} catch (err) {
  _iterator.e(err);
} finally {
  _iterator.f();
}
```

With the scopes proposal, is it possible to somehow mark `_loop` so that it does not appear in the stack when placing a breakpoint at `console.trace(x)`?

jridgewell:
I agree with @szuend, that's a generated range without an original scope. This should already be implemented in my beta `scopes` branch.

szuend:
I didn't implement anything yet but just thinking out loud. I modified the example a bit so we have the following input:

```js
function run(x) {
  x();
}

function foo() {
  for (let x of [1, 2, 3]) {
    console.trace(x);
    run(() => x);
  }
}

foo();
```

And Babel generates:

```js
function run(x) {
  x();
}
function foo() {
  var _loop = function _loop() {
    var x = _arr[_i];
    console.trace(x);
    run(function () {
      return x;
    });
  };
  for (var _i = 0, _arr = [1, 2, 3]; _i < _arr.length; _i++) {
    _loop();
  }
}
foo();
```

Now I would expect the following generated ranges (among others): One for `foo`, that points back the the original `foo` function scope. A generated range for

```js
    var x = _arr[_i];
    console.trace(x);
    run(function () {
      return x;
    });
```

that points back to the original block scope for the `for` loop. And a generated range for the `_loop();` call site (or the whole generated `for` loop body) with no original scope definition.

The stack trace for pausing on the `console.trace` line would be:

```
_loop (script.js 7:5)
foo (script.js 13:5)
<anonymous> (script.js 17:1)
```

[...]

PR #114, opened 2024-07-08
Scopes: Re-purpose `isScope` as `isFunctionScope` for stack trace processing

szuend:
This PR renames `isScope` to `isFunctionScope`. The intention is to mark each generated range that is also something callable (functions, methods, arrow functions, accessors, constructors, ...). The purpose is twofold:

1. Stack trace processors can use this to figure out which call frames to show. The check basically becomes `range.isFunctionScope && !range.hasDefinition`. See #113 for a detailed discussions and also why a `isHidden` flag is not sufficient.
1. It makes inlining easier to figure out for stack trace processors: With this flag they can tell exactly at which generated range to stop expanding inlined frames.

Note that for 1. in particular this is a great boon for generators. If we don't have this flag then generators have to "mask" callsites for "hidden functions" instead. Which is a lot trickier especially if not all call-sites are known by a generator. Again, see #113 for an example/discussion.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -136,7 +136,11 @@ interface OriginalScope {
 interface GeneratedRange {
   start: GeneratedPosition;
   end: GeneratedPosition;
-  isScope: boolean;
+  /**
+   * If the generated range is a function/method/accessor in the generated JavaScript code (anything callable).
+   * Consumers use this to figure out which call frames to include/exclude when processing stack traces.
+   */
+  isFunctionScope: boolean;
   originalScope?: OriginalScope;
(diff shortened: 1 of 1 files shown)


jkup:
@hbenl @takikawa @jridgewell Any thoughts on this one? I wasn't sure if it was blocking progress on the Scopes proposal implementations or not!

jridgewell:
The change seems fine, but I don't really have an opinion one way or the other because it doesn't affect the build-tooling side (only the consuming side). If @szuend thinks this makes consuming the ranges easier for dev-tools, then let's merge.

szuend:
This was merged as #147.

PR #117, opened 2024-07-29
Add examples for Base64 VLQ encoding

takikawa:
This PR adds some examples to the spec document for VLQ encoding.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -148,7 +148,8 @@ The string `"iB"` represents a [=Base64 VLQ=] with two digits. The first digit
 `"i"` encodes the bit pattern `0x100010`, which has a continuation bit of `1` (the
 VLQ continues), a sign bit of `0` (non-negative), and the value bits `0x0001`.
 The second digit `B` encodes the bit pattern `0x000001`, which has a continuation
-bit of `0` and no sign bit. The decoding of this VLQ string is the number 17.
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -143,6 +143,21 @@ Note: The values that can be represented by the VLQ Base64 encoded are limited t
 means that values exceeding 32-bits are invalid and implementations may reject
 them. The sign bit is counted towards the limit, but the continuation bits are not.
 
+<div class="example">
+The string `"iB"` represents a [=Base64 VLQ=] with two digits. The first digit
+`"i"` encodes the bit pattern `0x100010`, which has a continuation bit of `1` (the
+VLQ continues), a sign bit of `0` (non-negative), and the value bits `0x0001`.
+The second digit `B` encodes the bit pattern `0x000001`, which has a continuation
+bit of `0` and no sign bit. The decoding of this VLQ string is the number 17.
```
nicolo-ribaudo, line 151:
```suggestion
bit of `0`, no sign bit, and value bits `0x00001`. The decoding of this VLQ string is the number 17.
```

nicolo-ribaudo (approved):
Thanks! Let's clarify that `B` has 5 bits of value in that example.

Having these examples is helpful :)

takikawa:
Thanks for the suggestion! Should be included now.

jkup (approved):

PR #118, opened 2024-07-30
Update the scopes proposal example for recent changes

takikawa:
I noticed that the main example in the scopes proposal document seems to be out-of-date with the current data representation/encoding details. This PR is an attempt to update it.

I'd appreciate any feedback on the content because I'm not sure if I've used the fields in the intended way or if the scopes look quite right.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -358,7 +358,7 @@ End Generated Range C1 // B
 Start Generated Range C0 { // C
   field flags: has definition, has callsite
   definition: file.js, scope offset 1
-  callsite: file.js L6 C0
+  callsite: file.js L5 C0
   bindings: message -> "Hello World", y -> 2
 }
 End Scope C28 // C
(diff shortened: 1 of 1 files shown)


proposals/scopes.md
```diff
@@ -307,38 +337,35 @@ A|    var _x = 1;
...
 ;
-Start Scope C0 { // C
-  field flags: has name, has definition, has callsite
-  info flags: function, inherit parent bindings
-  name: z
-  definition: file.js L2 C0 - L5 C1
+Start Generated Range C0 { // C
+  field flags: has definition, has callsite
+  definition: file.js, scope offset 1
   callsite: file.js L6 C0
```
szuend, line 361: Even though this PR doesn't change this, but shouldn't the call-site be `L5 C0` (assuming 0-based numbers)?

takikawa: Thanks for catching that, should be updated now.

jkup (approved):

PR #119, opened 2024-08-08
Define decoding algorithms for source maps

nicolo-ribaudo:
### Preview: https://nicolo-ribaudo.github.io/source-map/#decoded-source-map

This PR introduces "decoded source map" data structures, which are
internal spec representations of the information encoded in source maps.

It also defines algorithms to decode source maps from either a JSON
string or [infra](https://infra.spec.whatwg.org/) representation.

The goal is:
- use them to explicitly write down all the possible error cases
- use them as a starting point to define new data structures, for
  example for the scopes proposal
- eventually add algorithms such as "get the original location given a
  decoded source map and a generated location".

This PR also explicitly defines sources/sourceRoot resolution in terms
of the [WHATWG URL](https://url.spec.whatwg.org/) spec.

There are a few points where I'm not sure about what the current/expected behavior is; I marked them with HTML comments (search for `<!--`). @takikawa Did you happen to figure out an answer to them while writing tests?
Also, I almost everywhere used "optionally report an error", but in a few cases I wrote "throw an error" because I don't think there is a way to recover from it (for example, invalid data in the `mappings` string). I didn't test what implementations do though.

This PR partially fixes the first and last points of https://github.com/tc39/source-map/issues/105. Closes https://github.com/tc39/source-map/issues/123.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -490,9 +490,6 @@ position, run the following stepsis the number :
 1. Advance |position| by 1.
 1. Return |value| * |sign|.
 
- return the result of decoding a base64 VLQ from |segment| starting at position |position|,
-and increase |position| by the number of bytes consumed. Return null if it's not possible to do so.
-
 NOTE: In addition to returning the decoded value, this algorithm updates the [=position variable=]
 in the calling algorithm.
 
--- a/source-map.bs
+++ b/source-map.bs
@@ -438,15 +438,15 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
             1. [=Decode a base64 VLQ=] from |segment| given |position| and let
                 |relativeOriginalLine| be the result.
             1. [=Decode a base64 VLQ=] from |segment| given |position| and let
-                |relativeOriginalPosition| be the result.
-            1. If |relativeOriginalPosition| is null, then:
+                |relativeOriginalColumn| be the result.
+            1. If |relativeOriginalColumn| is null, then:
                 1. If |relativeSourceIndex| is not null, [=optionally report an error=].
                 1. Continue with the next iteration.
             <!-- TODO: If there is a relativeSourceIndex but not a relativeOriginalLine,
                  should |sourceIndex| still be increased? -->
             1. Increase |sourceIndex| by |relativeSourceIndex|.
             1. Increase |originalLine| by |relativeOriginalLine|.
-            1. Increase |originalColumn| by |relativeOriginalPosition|.
+            1. Increase |originalColumn| by |relativeOriginalColumn|.
             1. If any of |sourceIndex|, |originalLine|, or |originalColumn| are less than 0, or if
                 |sourceIndex| is greater than or equal to |sources|'s [=list/size=], [=optionally
                 report an error=].
--- a/source-map.bs
+++ b/source-map.bs
@@ -410,6 +410,7 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
 1. Let |groups| be the result of [=strictly split|strictly splitting=] |mappings| on `;`.
 1. Let |generatedLine| be 0.
 1. Let |sourceIndex| be 0.
+1. Let |nameIndex| be 0.
 1. While |generatedLine| is less than |groups|'s [=list/size=]:
     1. If |groups|[|generatedLine|] is not the empty string, then:
         1. Let |segments| be the result of [=strictly split|strictly splitting=]
@@ -417,7 +418,6 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
         1. Let |generatedColumn| be 0.
         1. Let |originalLine| be 0.
         1. Let |originalColumn| be 0.
-        1. Let |nameIndex| be 0.
         1. [=For each=] |segment| in |segments|:
             1. Let |position| be a [=position variable=] for |segment|, initially pointing at
                 |segment|'s start.
--- a/source-map.bs
+++ b/source-map.bs
@@ -404,7 +404,7 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
     NOTE: These are the valid [[base64]] characters (excluding the padding character `=`), together
     with `,` and `;`.
 
-    then, throw an error.
+    then throw an error.
 1. Let |decodedMappings| be a new empty [=list=].
 1. Let |groups| be the result of [=strictly split|strictly splitting=] |mappings| on `;`.
 1. Let |generatedLine| be 0.
@@ -459,10 +459,10 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
             1. [=Decode a base64 VLQ=] from |segment| given |position| and let |relativeNameIndex|
                 be the result.
             1. If |relativeNameIndex| is not null, then:
-              1. Increase |nameIndex| by |relativeNameIndex|.
-              1. If |nameIndex| is negative or greater than |names|'s [=list/size=], [=optionally
-                  report an error=].
-              1. Else, set |decodedMapping|'s [=decoded mapping/name=] to |names|[|nameIndex|].
+                1. Increase |nameIndex| by |relativeNameIndex|.
+                1. If |nameIndex| is negative or greater than |names|'s [=list/size=], [=optionally
+                    report an error=].
+                1. Else, set |decodedMapping|'s [=decoded mapping/name=] to |names|[|nameIndex|].
             1. If |position| does not point to the end of |segment|, [=optionally report an
                 error=].
     1. Increase |generatedLine| by 1.
--- a/source-map.bs
+++ b/source-map.bs
@@ -399,7 +399,6 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
     - U+0041 (A) to U+005A (Z);
     - U+0061 (a) to U+007A (z);
     - U+002B (+), U+002F (/)
-    <!-- TODO: Is = allowed? -->
 
     NOTE: These are the valid [[base64]] characters (excluding the padding character `=`), together
     with `,` and `;`.
@@ -408,12 +407,12 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
 1. Let |decodedMappings| be a new empty [=list=].
 1. Let |groups| be the result of [=strictly split|strictly splitting=] |mappings| on `;`.
 1. Let |generatedLine| be 0.
+1. Let |sourceIndex| be 0.
 1. While |generatedLine| is less than |groups|'s [=list/size=]:
     1. If |groups|[|generatedLine|] is not the empty string, then:
         1. Let |segments| be the result of [=strictly split|strictly splitting=]
             |groups|[|generatedLine|] on `,`.
         1. Let |generatedColumn| be 0.
-        1. Let |sourceIndex| be 0.
         1. Let |originalLine| be 0.
         1. Let |originalColumn| be 0.
         1. Let |nameIndex| be 0.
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+<dfn for="decode source map mappings">|names|</dfn>, and a [=list=] of [=decoded source|decoded
+sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the following steps:
+1. If |mappings| is not an [=ASCII string=], throw an error.
+1. If |mappings| contains any [=code unit=] other than:
+    - U+002C (,) or U+003B (;);
+    - U+0030 (0) to U+0039 (9);
+    - U+0041 (A) to U+005A (Z);
+    - U+0061 (a) to U+007A (z);
+    - U+002B (+), U+002F (/)
+    <!-- TODO: Is = allowed? -->
```
takikawa, line 402: I don't think implementations allow `=`. e.g., [Chrome](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/common/Base64.ts;l=5;drc=82dff63dbf9db05e9274e11d9128af7b9f51ceaa) and [Mozilla's library](https://github.com/tromey/vlq/blob/master/src/lib.rs#L33). But it seems worth clarifying that in the spec. I also don't think I wrote any tests for that case, so I'll add some.

source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+    then throw an error.
+1. Let |decodedMappings| be a new empty [=list=].
+1. Let |groups| be the result of [=strictly split|strictly splitting=] |mappings| on `;`.
+1. Let |generatedLine| be 0.
+1. While |generatedLine| is less than |groups|'s [=list/size=]:
+    1. If |groups|[|generatedLine|] is not the empty string, then:
+        1. Let |segments| be the result of [=strictly split|strictly splitting=]
+            |groups|[|generatedLine|] on `,`.
+        1. Let |generatedColumn| be 0.
+        1. Let |sourceIndex| be 0.
```
takikawa, line 416: Should `|sourceIndex|` through `|nameIndex|` be defined outside of this loop? I believe these should retain their previous value even when the `generatedLine` goes to the next line.

source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+            1. Increase |originalColumn| by |relativeOriginalPosition|.
+            1. If any of |sourceIndex|, |originalLine|, or |originalColumn| are less than 0, or if
+                |sourceIndex| is greater than or equal to |sources|'s [=list/size=], [=optionally
+                report an error=].
+            1. Else,
+                1. Set |decodedMapping|'s [=decoded mapping/originalSource=] to
+                    |sources|[|sourceIndex|].
+                1. Set |decodedMapping|'s [=decoded mapping/originalLine=] to |originalLine|.
+                1. Set |decodedMapping|'s [=decoded mapping/originalColumn=] to |originalColumn|.
+            1. [=Decode a base64 VLQ=] from |segment| given |position| and let |relativeNameIndex|
```
jridgewell, line 459: Before doing this, check for EOF/`,`/`;`, and continue if so.

source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+    1. Advance |position| by 1.
+    1. If |position| points to the end of |segment|, throw an error.
+    1. Set |currentByte| to the [=byte=] whose the [=byte/value=] is the number corresponding to
+        |segment|'s |position|th [=code unit=], according to the [[base64]] encoding.
+    1. Let |chunk| be |currentByte| & 0x1F, as a number.
+    1. Add |chunk| &lt;&lt; |nextShift| to |value|.
+    1. If |value| is greater than or equal to 2<sup>31</sup>, throw an error.
+    1. Increase |nextShift| by 5.
+1. Advance |position| by 1.
+1. Return |value| * |sign|.
```
jridgewell, line 491: There's a bug here when decoding `B` (which corresponds to -2147483648):
- `first = 1`
- `currentByte = 1`
- `value = 0`
- `sign = -1`
- `value * sign === -0`, not `-2147483648`

source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+
+NOTE: In addition to returning the decoded value, this algorithm updates the [=position variable=]
```
jridgewell, line 493: Can we just return a tuple?

source-map.bs
```diff
@@ -185,37 +185,145 @@ following structure:
...
+1. Let |jsonMap| be the result of [=parse a JSON string to an Infra value|parsing a JSON string to
+    an Infra value=] |str|.
+1. If |jsonMap| is not a [=map=], report an error and abort these steps.
+1. [=Decode a source map=] given |jsonMap| and |baseURL|, and return its result if any.
+
+To <dfn export>decode a source map</dfn> given a [=string=]-keyed [=map=] |jsonMap| and a [=/URL=]
+|baseURL|, run the following steps:
+1. If |jsonMap|[`"version"`] does not [=map/exist=] or |jsonMap|[`"version"`] is not 3,
+    [=optionally report an error=].
+1. If |jsonMap|[`"mappings"`] does not [=map/exist=] or |jsonMap|[`"mappings"`], is not a
```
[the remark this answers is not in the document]
nicolo-ribaudo: The PR used to fallback to an empty array just because it's easy to do so, but indeed it looks like both Chrome and Firefox error in this case.

source-map.bs
nicolo-ribaudo: This is based on an existing pattern in other bikeshed/web specs to use this "dynamic variable", defined at https://infra.spec.whatwg.org/#string-position-variable

source-map.bs
```diff
@@ -256,14 +364,166 @@ with four fields represent mapped code where a corresponding name does not exist
...
+                [=optionally report an error=] and continue with the next iteration.
+            1. Let |decodedMapping| be a new [=decoded mapping=] whose
+                [=decoded mapping/generatedLine=] is |generatedLine|,
+                [=decoded mapping/generatedColumn=] is |generatedColumn|,
+                [=decoded mapping/originalSource=] is null,
+                [=decoded mapping/originalLine=] is null,
+                [=decoded mapping/originalColumn=] is null,
+                and [=decoded mapping/name=] is null.
+            1. Append |decodedMapping| to |decodedMappings|.
+            1. [=Decode a base64 VLQ=] from |segment| given |position| and let |relativeSourceIndex|
```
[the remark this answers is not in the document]
nicolo-ribaudo: `segment[position]` cannot be `,`/`;`, because segment is obtained by first splitting the string on those two chars. I'll handle the end-of-string case.

source-map.bs
nicolo-ribaudo: Chrome's implementation also decodes `B` to 0 (it returns -0, since it uses floats): https://github.com/ChromeDevTools/devtools-frontend/blob/9b580b72f1334bb91778a052b0bc4bdc1eb8c452/front_end/core/sdk/SourceMap.ts#L846

nicolo-ribaudo: Reading https://tc39.es/source-map/#base64-vlq:

> Base64 VLQ: [[VLQ]](https://tc39.es/source-map/#biblio-vlq) is a [[base64]](https://tc39.es/source-map/#biblio-base64) value, where the most significant bit (the 6th bit) is used as the continuation bit, and the "digits" are encoded into the string least significant first, and where the least significant bit of the first digit is used as the sign bit.

It seems to me that, when decoding B (000001):
- the continuation bit is 0 (so this VLQ is complete)
- the sign bit is 1
- the digits are 0000

0000 is 0, so the absolute value of this number is 0.

jridgewell: I did fix Chrome years ago: https://chromium-review.googlesource.com/c/v8/v8/+/2018882

nicolo-ribaudo: I opened https://github.com/tc39/source-map/issues/123 to continue the normative discussion, for now I'll update this PR to let implementations return _either_ 0 or -2147483648. My goal with this PR is to clarify what is already happening and what the spec already says, and not to force changes in implementations.

nicolo-ribaudo:
PR updated to return i32::min for B.

source-map.bs
```diff
@@ -410,14 +410,14 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
 1. Let |groups| be the result of [=strictly split|strictly splitting=] |mappings| on `;`.
 1. Let |generatedLine| be 0.
 1. Let |sourceIndex| be 0.
+1. Let |nameIndex| be 0.
 1. While |generatedLine| is less than |groups|'s [=list/size=]:
     1. If |groups|[|generatedLine|] is not the empty string, then:
         1. Let |segments| be the result of [=strictly split|strictly splitting=]
             |groups|[|generatedLine|] on `,`.
         1. Let |generatedColumn| be 0.
         1. Let |originalLine| be 0.
```
jridgewell, line 419: Hoist this one, too.

source-map.bs
```diff
@@ -410,14 +410,14 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
...
         1. Let |originalColumn| be 0.
```
jridgewell, line 420: Hoist this one, too.

source-map.bs
```diff
@@ -256,14 +366,172 @@ with four fields represent mapped code where a corresponding name does not exist
...
+    1. Let |decodedSource| be a new [=decoded source=] whose [=decoded source/URL=] is null,
+        [=decoded source/content=] is null, and [=decoded source/ignored=] is false.
+    1. If |source| is not null:
+        1. Set |source| to the [=string/concatenate|concatenation=] of |sourceURLPrefix| and |source|.
+        1. Let |sourceURL| be the result of [=URL parser|URL parsing=] |source| with |baseURL|.
+        1. If |sourceURL| is failure, [=optionally report an error=].
+        1. Else, set |decodedSource|'s [=decoded source/URL=] to |sourceURL|.
+    1. If |index| is in |ignoredSources|, set |decodedSource|'s [=decoded source/ignored=] to true.
+    1. If |sourcesContent|'s [=list/size=] is greater than or equal to |index|, set
+        |decodedSource|'s [=decoded source/content=] to |sourcesContent|[|index|].
```
takikawa, line 531: The handling of sourcesContent here seems like a natural way to spec it, but when you have duplicate entries it currently conflicts with what implementations actually do. (see [this issue](https://github.com/tc39/source-map/issues/85) for more details) In particular, it seems all the major browsers treat sources entries as being set-like so that duplicate source URLs point to the same sourcesContent (either the first or last one for that particular URL).

jridgewell (approved):

source-map.bs
```diff
@@ -256,14 +366,176 @@ with four fields represent mapped code where a corresponding name does not exist
...
+                be the result.
+            1. [=Decode a base64 VLQ=] from |segment| given |position| and let
+                |relativeOriginalLine| be the result.
+            1. [=Decode a base64 VLQ=] from |segment| given |position| and let
+                |relativeOriginalColumn| be the result.
+            1. If |relativeOriginalColumn| is null, then:
+                1. If |relativeSourceIndex| is not null, [=optionally report an error=].
+                1. Continue with the next iteration.
+            <!-- TODO: If there is a relativeSourceIndex but not a relativeOriginalLine,
+                 should |sourceIndex| still be increased? -->
```
takikawa, line 448: I think this TODO can be deleted as the answer to this should be "no" right? (because if relativeOriginalLine is null, then relativeOriginalColumn must also be null due to where VLQ decoding returns null, and thus it optionally returns an error)

takikawa (approved):
LGTM, the new note about sources looks good thanks. Left a minor comment about a remaining TODO.

source-map.bs
nicolo-ribaudo, line 448:
```suggestion
```

PR #122, opened 2024-08-26
[names] Add a section that clarifies how to emit names for JS

szuend:
This is a first draft that clarifies on which JS tokens we expect mappings with a `name`.

I tried to stick to the ECMA-262 AST on a level where it makes sense, but let me know what you think. I also used lang

A nice tie-in with the "Scopes" proposal is, that the `(` token that shows up in the definition is also what we'll probably recommend as the "generated range" start, so that would fit rather well together.

Thinks that still need clarification:
- [ ] Phrasing around the conditions "when" to emit a mapping with a `name`. Since we don't know the source language we need to keep it generic but we should make it clear that it should also make sense semantically.
- [ ] The current phrasing uses "should" and "may". I'm not sure if we should phrase it as "must"?
- [ ] Should we spell out all the AST constructs for variables where we expect mappings with a `name`?
- [ ] Figure out if we want to allow more complex expressions for variable names, e.g. if a minifier turned `obj.x` into `f`.
- [ ] Any other open question you might come up with :)


source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+
+1. The opening parenthesis `(` of [=MethodDefinition=] for class methods. For the
+    [=MethodDefinition=] where *PropName* is "constructor", the [=name=] should be
+    the original class name.
+
+1. The opening parenthesis `(` of [=GeneratorDeclaration=], [=GeneratorExpression=],
+    [=AsyncGeneratorDeclaration=] and [=AsyncGeneratorExpression=] for (async) generator functions.
+
+1. The first letter of [=Identifier=] for variables. This includes (but is not limited to)
+    [=LexicalDeclaration=], [=VariableStatement=] and arguments in [=Parameter List=].
```
nicolo-ribaudo, line 323: Is "function declaration" in this list?

e.g.

```js
var foo = function () {}
```
```js
function foo() {}
```

to they both have two name mappings? (one for the variable they introduce, and one for the actual function)

source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+    Source map generators may choose to only emit a [=mapping=] entry with a [=name=] at a
+    variable's definition site and may choose to omit such a mapping for further use-sites
+    in the same [=Environment Record=].
```
nicolo-ribaudo, line 326: `Environment Record` is a _runtime_ concept, so it doesn't exist at parse/analysis time: maybe we should say "in the same scope"? Also, in
```js
function fn() {
  let x;
  x()
  function nested() { x() }
}
```
the two references to `x` are in two separate environment records. Do we need the "same environment record" restriction?

szuend: Alternatively: We can also remove the distinction between definition site and use-site and recommend emitting name mappings generally for all `BindingIdentifier` in all expressions (so also all use-sites as well).

source-map.bs
szuend: Isn't the first example a `VariableStatement` plus `FunctionExpression`?

For the second case, yes, we could recommend two name mappings, one on the identifier and one on the paren.

source-map.bs
szuend: Excellent point. I'm not that familiar with the official spec AST (yet).

So we could basically spec something like: "generators SHOULD emit a name mapping for `BindingIdentifier` and they MAY emit a name mapping for `IdentifierReference`?

sjrd: Yes, something like seems reasonable.

We could also recommend *not* to emit name mappings for `IdentifierReference`s, mostly because that would be bytes spent for no reason?

source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+
+1. The original source language construct maps semantically to the generated
+    JavaScript code.
+
+1. The original source language construct has a name.
+
+Then the [=name=] of the [=mapping=] entry should be the name of the original
+source language construct.
+
+Example: A TypeScript arrow function compiled to a JavaScript function, or a
```
rbuckton, line 306: Is an arrow function a good example? Arrow functions don't have names unless they are assigned names, i.e., `let f = () => {}`. It isn't clear to me if this section intends for a source map generator to emit the name `f` for that function.

source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+local variable that was renamed.
+
+1. The opening parenthesis `(` of [=FunctionDeclaration=], [=FunctionExpression=],
```
rbuckton, line 309: Does this mean that the generator should add a mapping with a name pointing to `(`, even if the generated function also has a name? That would map the `name` to the token that follows the actual name. This seems to conflict with (5) where variables and parameter names are mapped to the first character of the identifier.

source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+    [=AsyncFunctionDeclaration=] and [=AsyncFunctionExpression=] for (async) functions.
+
+1. The opening parenthesis `(` or the first letter of the *BindingExpression*
```
rbuckton, line 312: Wouldn't this conflict with (5), since `x => {}` could potentially require `name` mappings both for the arrow function and for the parameter name. It would almost be better to map the `name` to the `=>` token, since that is unambiguously part of the arrow function itself.

source-map.bs
rbuckton, line 323: I would prefer wording that does not conflate "arguments" with "parameters", as "arguments" generally refers to the values you supply when performing invocation, while "parameters" refers to the names to which those values are bound within the function. 

```suggestion
    [=LexicalDeclaration=], [=VariableStatement=] and [=Parameter List=].
```

I would suggest just removing the wording.

source-map.bs
```diff
@@ -274,6 +290,44 @@ fields represent mapped code that also has a mapped name.
...
+1. The opening parenthesis `(` of [=GeneratorDeclaration=], [=GeneratorExpression=],
+    [=AsyncGeneratorDeclaration=] and [=AsyncGeneratorExpression=] for (async) generator functions.
+
+1. The first letter of [=Identifier=] for variables. This includes (but is not limited to)
+    [=LexicalDeclaration=], [=VariableStatement=] and arguments in [=Parameter List=].
+    Source map generators may choose to only emit a [=mapping=] entry with a [=name=] at a
+    variable's definition site and may choose to omit such a mapping for further use-sites
+    in the same [=Environment Record=].
+
+Note: The opening parenthesis is chosen because it is present in all styles of function
```
rbuckton, line 328: As I stated above, it may be worth considering having the name mapped to the `=>` of a paren-less arrow with single parameter to avoid ambiguity with a mapping for the parameter itself.

source-map.bs
nicolo-ribaudo: The function in this case has name `f`. You can test it by running
```js
let f = () => {};
console.log(f.name);
```

rbuckton: > The function in this case has name `f`. You can test it by running

Yes, I mentioned assigned names in my initial comment. To clarify, I'm asking if this section should more clearly spell out that a name could be defined for an arrow function for cases such as this.

szuend: Yeah I think the other way makes more sense as an example: "A TypeScript function compiled to a JavaScript arrow function, ..."

source-map.bs
szuend: Yes, for function/generator declarations with a `BindingIdentifier` in the generated code we basically recommend emitting a named mapping for both the `BindingIdentifier` as well as the `(`.

The advantage is that for consumers this is somewhat easier since they don't have to distinguish too many cases and encoding wise it's not too expensive as the name on the `(` will always be encoded as `A` (`0`) since it's the same name.

source-map.bs
szuend: Great point. Should we carve out an exception for paren-less, single-argument arrow functions or generally recommend to put the `name` on the `=>` token for all arrow functions?

I'm leaning to the former but an argument could also be made for the latter given that all arrow functions could be handled identically if the `=>` mapping contains the name.

source-map.bs
szuend: Agreed, I'll remove `arguments in ` in the next commit.

source-map.bs
szuend: Done

source-map.bs
szuend: I updated the text to read that generators "should" emit  the name on the mapping for `BindingIdentifier` or on the opening paren if no identifier exists. But they may emit it on the opening paren regardless.

szuend:
Hey all, I incorporated the feedback form yesterday's meeting. Thanks a ton for the fruitful discussion! In particular:

  * For functions/methods/generators emit the named mapping on the
    binding identifier by default (or opening parenthesis if no identifier is present).

  * Special case single argument arrow functions that have neither
    a binding identifier or an opening parenthesis.

  * Introduce a definition for "named mapping". I.e. a decoded
    mapping entry where the name field is not null.

  * Call out that generators may emit additional "named mappings"
    as they require. Either to support existing tools or to fit
    their requirements.

English is not my native language and writing spec text is not my strong suite, so I'd welcome some word smithing/rephrasing to clear up the meaning. I also spelled out some more things that require review. Thanks!

source-map.bs
```diff
@@ -590,6 +611,65 @@ as per [=ECMAScript Lexical Grammar=]. [=Mapping=] entries must point to either:
...
+1. The original source language construct maps semantically to the generated
+    JavaScript code.
+
+1. The original source language construct has a name.
+
+Then the [=decoded mapping/name=] of the [=decoded mapping|mapping=] entry should be
+the name of the original source language construct. A [=decoded mapping|mapping=] with
+a non-null [=decoded mapping/name=] is called a <dfn>named mapping</dfn>.
+
+Example: A TypeScript function compiled to a JavaScript arrow function, or a
```
rbuckton, line 628: In general, TypeScript does not transpile a `function` to an arrow function in any case. Instead, we might transpile an arrow function with type information to an arrow function without, or an arrow function to a `function` as part of down-leveling to ES5.

source-map.bs
```diff
@@ -590,6 +611,65 @@ as per [=ECMAScript Lexical Grammar=]. [=Mapping=] entries must point to either:
...
+
+1. The [=BindingIdentifier=] for [=FunctionDeclaration=], [=FunctionExpression=],
+    [=AsyncFunctionDeclaration=], [=AsyncFunctionExpression=], [=GeneratorDeclaration=],
+    [=GeneratorExpression=], [=AsyncGeneratorDeclaration=], and [=AsyncGeneratorExpression=] if it
+    exists, or the opening parenthesis `(` preceding the [=FormalParameters=] otherwise.
+
+    Source map generators may chose to emit a [=named mapping=] on the opening parenthesis regardless of
+    the presence of the [=BindingIdentifier=].
+
+1. The [=BindingIdentifier=] for [=ArrowFunction=], and [=AsyncArrowFunction=] if it
```
rbuckton, line 656: This paragraph is confusing and doesn't properly address the concern I raised. It now says to emit a name mapping on (a) either the _BindingIdentifier_ or `(` of an arrow function, or (b) the `=>` if it is a single argument arrow function with neither a _BindingIdentifier_ nor _Arguments_.

(b) is never true as a single argument arrow function has _only_ a _BindingIdentifier_, and thus (a) is only the ever case that is true.

Correct me if I'm wrong, but I believe we discussed in the meeting that there are two possible places to emit a name mapping for an arrow function or async arrow function:

1. For an _ArrowFunction_ or _AsyncArrowFunction_ with _Arguments_, the name mapping is attached to the opening `(`. A generator _may_ opt to emit a name mapping attached to the `=>` for consistency with (2).
2. For an _ArrowFunction_ or _AsyncArrowFunction_ with a _SingleNameBinding_, the name mapping is attached to the `=>`.

source-map.bs
szuend: Replaced with "A minifier renaming functions and variables or removing function names
from immediately invoked function expressions. "

source-map.bs
szuend: I'm embarrassed to say that I don't know what I was thinking when I wrote this.

I fixed it up as proposed but with concrete references to the productions.

rbuckton (approved):

PR #124, opened 2024-08-28
[mappings] Define for each ECMAScript input element where mappings are allowed

szuend:
This PR is a initial take to at least clarify on a very basic level where we allow mappings for JavaScript. The current draft proposes to use "input elements" instead of "ECMAScript tokens" since generators might want to map comments as well (e.g. a TypeScript -> JavaScript with JSDoc transformer that maps the `@param` back to the TS type annotation).

The main reasoning behind the current draft is that it seldom make sense to have mappings in the middle of tokens. E.g. with `let foo = 5;`, you don't really need a mapping for the `o` in `foo`. The exception (I think) are string literals where minifiers could combine multiline string literals into a single literal. Template literals might also fall under this category so we could move them down to the second bullet point.

I don't see a use case for mapping white space and line terminators, but if there is a valid use-case we should also move them to the second bullet point.

Note: I plan to do a followup where we recommend some AST productions where generators "should" emit mappings to better facilitate stack trace deobfuscation/debugging/etc. But this is kinda tricky since engines don't really agree on where `Error.stack` frames point to w.r.t. to the ECMAScript AST, so we'd have to cover all or at least the major use cases in the major browsers/runtimes.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -43,6 +43,21 @@ urlPrefix:https://tc39.es/ecma262/#; type:dfn; spec:ecmascript
     url:prod-MultiLineComment; text:multi-line comment
     url:prod-MultiLineComment; text:multi-line comment
     url:sec-regexpbuiltinexec; text:RegExpBuiltinExec
+    url:prod-Comment; text:Comment
+    url:prod-DivPunctuator; text:DivPunctuator
+    url:prod-HashbangComment; text:HashbangComment
+    url:prod-IdentifierName; text:IdentifierName
+    url:prod-LineTerminator; text:LineTerminator
+    url:prod-NumericLiteral; text:NumericLiteral
+    url:prod-Punctuator; text:Punctuator
+    url:prod-PrivateIdentifier; text:PrivateIdentifier
+    url:prod-RegularExpressionLiteral; text:RegularExpressionLiteral
+    url:prod-RightBracePunctuator; text:RightBracePunctuator
+    url:prod-StringLiteral; text:StringLiteral
+    url:prod-Template; text:Template
+    url:prod-TemplateSubstitutionTail; text:TemplateSubstitutionTail
+    url:prod-WhiteSpace; text:WhiteSpace
+    url:sec-ecmascript-language-lexical-grammar; text:ECMAScript Lexical Grammar
 
 urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
     url:binary/modules.html#binary-customsec; text:custom section
@@ -273,6 +288,21 @@ fields represent mapped code that also has a mapped name.
 
 Note: Using file offsets was considered but rejected in favor of using line/column data to avoid becoming
 misaligned with the original due to platform-specific line endings. 
+
+### [=Mappings=] for generated JavaScript code
+
+Generated code positions that may have [=Mapping=] entries are defined in terms of *input elements*
+as per [=ECMAScript Lexical Grammar=]:
+
+1. [=Mapping=] entries must point to the first code point of the source text matched by
+    [=IdentifierName=], [=PrivateIdentifier=], [=Punctuator=], [=DivPunctuator=], [=RightBracePunctuator=],
+    [=NumericLiteral=], [=RegularExpressionLiteral=], [=Template=] and [=TemplateSubstitutionTail=].
+
+1. [=Mapping=] entries may point to any code point of the source text matched by
+    [=Comment=], [=HashbangComment=] and [=StringLiteral=].
+
+1. [=Mapping=] entries must not point to any code point of the source text matched by
+    [=WhiteSpace=] and [=LineTerminator=].
 
 Resolving Sources {#resolving-sources}
 --------------------------------------
--- a/source-map.bs
+++ b/source-map.bs
@@ -296,13 +296,11 @@ as per [=ECMAScript Lexical Grammar=]:
 
 1. [=Mapping=] entries must point to the first code point of the source text matched by
     [=IdentifierName=], [=PrivateIdentifier=], [=Punctuator=], [=DivPunctuator=], [=RightBracePunctuator=],
-    [=NumericLiteral=], [=RegularExpressionLiteral=], [=Template=] and [=TemplateSubstitutionTail=].
+    [=NumericLiteral=] and [=RegularExpressionLiteral=].
 
 1. [=Mapping=] entries may point to any code point of the source text matched by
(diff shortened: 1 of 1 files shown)


sjrd:
Mappings in the generated source have range positions. Why are we restricting them to only point to the first code point? It seems more natural to point to the entire range that is produced.

For example, for an identifier in source that becomes an identifier in the target, you'd want the mapping to cover the entire identifier in the target.

Likewise, for an operation such as an addition in source `a + b`, the produced code could look like `(a + b) | 0` if it implements arithmetic modulo 0, and is pretty-printed. In that case, arguably, all the characters of that string, except the substrings `a` and `b`, should map to the source `+` operator. That should include whitespace, if only to reduce the number of required mappings.

Overall I'm not sure what the proposed restrictions are meant to achieve? Can the consumers benefit from those restrictions for some reason?

szuend:
I'm not sure I fully understand. Mappings are not ranges. They only translate a single point in generated code to a single point in authored code. Encoding overlapping generated source ranges is not possible. Generators/consumers might interpret that consecutive mappings form ranges, but that's not actually spec'ed.

For your example, you could encode it in a couple different ways. Either map the generated  `(` to the authored `a` and the next non-whitespace token after `0` to after `b`. Or just map the `a + b` with empty mappings for the parens and `| 0` or add mappings for `a `, `+ ` and `b)` individually.

The only thing the PR attempting to do, is that it might not make sense if you write `(foo + bar)` to have any mapping start in the middle of `foo`.

From a DevTools perspective the mapping for an expression ala `(a + b) | 0` wouldn't matter much. The only thing we'd be interested here is where to put breakpoints and that happens 90% on the statement level.

ehoogeveen-medweb:
Regarding whitespace or more specifically line terminators, what about minifiers replacing (whitespace including 1 or more) newlines with a `;` to avoid/mimic automatic semicolon insertion? The other case I can think of where line terminators are load bearing would be for ending single-line comments, but I don't know if there's something relevant to map there.

szuend:
If a minifier replaces whitespace with a `;` you would add a mapping from the `;` to the whitespace which is still allowed. The whitespace would be in authored code, not in the generated bundle.

The other way around is interesting though, but the question is, if there is tooling/use cases for which a mapping from a whitespace to a `;` would be relevant?

ehoogeveen-medweb:
Although it seems somewhat contrived, I guess the reverse use case would be unminifying or deobfuscating with a code formatter that omits semicolons by default. You could use a source map for highlighting the corresponding minified or obfuscated code and vice versa, though I suppose you could always use the reverse mapping instead.

ehoogeveen-medweb:
Although it seems somewhat contrived, I guess the reverse use case would be unminifying or deobfuscating with a code formatter that omits semicolons by default. You could use a source map for highlighting the corresponding minified or obfuscated code and vice versa (probably as a serialization format rather than directly), though I suppose you could always use the reverse mapping instead.

sjrd:
> I'm not sure I fully understand. Mappings are not ranges. They only translate a single point in generated code to a single point in authored code. Encoding overlapping generated source ranges is not possible. Generators/consumers might interpret that consecutive mappings form ranges, but that's not actually spec'ed.

My understanding had always been that a mapping starts at the given offset, and ends where the next mapping starts. Which makes them range-based.

If instead they truly are points and must point to a very specific code point, and there are several consecutive tokens that must all map to the same source position, you have to emit a lot more segments, bloating the source map file.

szuend:
This is exactly the reason why I want to have the discussion and attempt to nail this down a bit in the spec.

I don't propose emitting a mapping for each token. For your use case where multiple tokens in the generated source map point to the same location in authored code, it would be fine to emit one mapping at the start and one mapping at the end of the range in the generated code. If you then throw or pause somewhere inside that generated source range, tooling would be able to find the authored point by taking the closest mapping (forwards or backwards would both work in this case).

In that sense, yes you are right and we do indeed have ranges. But we can't assume that all mappings close the range that the previous mapping point opened (or is the start of a new range). This would require generators to emit an empty (i.e. unmapped) mapping to close ranges. Which some generators may do and others may not.

I'd love to also harden the rules here, and we might be able to do so as a follow-up.

The PR here is on a more fundamental level. It only clarifies where in the generated source, mapping points are allowed. Generators can still choose to emit a mapping point for every token, just for some or for none at all.

W.r.t. to whitespace, I can see that there are valid use cases, e.g. if you'd like to mark that generated code after some token is unmapped you'd probably want to put an empty mapping there.

szuend:
I relaxed the restrictions around whitespace and templates as discussed in the meeting.

rbuckton (approved):

nicolo-ribaudo (dismissed):

source-map.bs
```diff
@@ -274,6 +289,19 @@ fields represent mapped code that also has a mapped name.
...
+Generated code positions that may have [=Mapping=] entries are defined in terms of *input elements*
+as per [=ECMAScript Lexical Grammar=]:
+
+1. [=Mapping=] entries must point to the first code point of the source text matched by
+    [=IdentifierName=], [=PrivateIdentifier=], [=Punctuator=], [=DivPunctuator=], [=RightBracePunctuator=],
+    [=NumericLiteral=] and [=RegularExpressionLiteral=].
+
+1. [=Mapping=] entries may point to any code point of the source text matched by
+    [=Comment=], [=HashbangComment=], [=StringLiteral=], [=Template=],
+    [=TemplateSubstitutionTail=], [=WhiteSpace=] and [=LineTerminator=].
```
nicolo-ribaudo, line 303: I agree with the contents, but I found the wording confusing (if mappings _must_ point to the first code point of things in the first item, how can they also point somewhere else)?

What about this?
```suggestion
Generated code positions that may have [=Mapping=] entries are defined in terms of *input elements*
as per [=ECMAScript Lexical Grammar=]. [=Mapping=] entries must point to either:

1. the first code point of the source text matched by
    [=IdentifierName=], [=PrivateIdentifier=], [=Punctuator=], [=DivPunctuator=], [=RightBracePunctuator=],
    [=NumericLiteral=] and [=RegularExpressionLiteral=].

1. any code point of the source text matched by
    [=Comment=], [=HashbangComment=], [=StringLiteral=], [=Template=],
    [=TemplateSubstitutionTail=], [=WhiteSpace=] and [=LineTerminator=].
```

szuend: Thanks for the improved wording!

nicolo-ribaudo (approved):

PR #126, opened 2024-09-13
Rework spec intro&co to match Ecma rules

nicolo-ribaudo:
Ecma has some rules on how standards should be structured: [TOOLS-013-1.pdf](https://github.com/user-attachments/files/16994332/TOOLS-013-1.pdf)

This PR:
- Adds the Introduction to the TOC, but withuot a number (see 6.1.4)
- Makes "Scope" the first normative section of the spec, describing what the spec covers (see 6.2.1)
- Renamed "Terminology" to "Terms and definitions (see 6.3.1), and used the bikeshed syntax for definition lists.
- Similarly to how it's done in the ECMAScript spec, I added an "About this Specification" section at the beginning that basically says "the online document is the most up-to-date version".
- I moved the license at the end, like the ECMAScript spec does.


source-map.bs
```diff
@@ -97,19 +104,23 @@ urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
 }
 </pre>
 
-## License
+<h2 class="no-num no-toc no-ecma-publish">About this Specification</h2>
```
nicolo-ribaudo, line 107:
```suggestion
<h2 class="no-num no-toc">About this Specification</h2>
```

source-map.bs
```diff
@@ -2,20 +2,27 @@
...
 Repository: source-map/source-map-spec
-Abstract: A specification for mapping transpiled source code back to the original sources.  This specification is a living document and describes a hardened version of the Source Map v3 specification.
+Abstract: This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
+No Abstract: true
 Markup Shorthands: markdown yes
 Group: tc39
 </pre>
 
+<link rel="stylesheet" media="print" href="./ecma-style/ecma.css">
+
```
nicolo-ribaudo, line 25:
```suggestion
```

source-map.bs
```diff
@@ -2,16 +2,21 @@
...
+Test Suite: https://github.com/tc39/source-map-tests
+Editor Term: Editor, Editors
 Editor: Asumu Takikawa, Igalia
 Former Editor: Victor Porof, Google
 Former Editor: John Lenz, Google
 Former Editor: Nick Fitzgerald, Mozilla
 Previous Version: https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1#
 Repository: source-map/source-map-spec
-Abstract: A specification for mapping transpiled source code back to the original sources.  This specification is a living document and describes a hardened version of the Source Map v3 specification.
+Abstract: This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
```
jkup, line 18: Curious if we'll get any pushback on this.

source-map.bs
```diff
@@ -2,16 +2,21 @@
...
+No Abstract: true
```
jkup, line 19: Just unfamiliar with this, why no abstract is set to true when we now have an abstract?

source-map.bs
```diff
@@ -97,19 +102,23 @@ urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
...
+Repository](https://github.com/tc39/source-map).
 
-## Introduction
+<h2 class="no-num">Introduction</h2>
 
-This document is a draft version of a hardened version of the Source Map v3
-specification.  In its current form, it's not a defined standard and is subject to
-modifications.  If you want to get involved you will find more information under
-the following [Specification Repository](https://github.com/tc39/source-map).
+This Ecma standard defines the Source Map format, used for mapping transpiled source code (primarily
```
jkup, line 116: Is "primarily JavaScript" helpful here?

jkup (commented):
Thank you for doing this! I just had a few quick questions.

source-map.bs
nicolo-ribaudo: On what part specifically? :)

source-map.bs
nicolo-ribaudo: Because in that PDF with rules about not having a standard it says to not have an "Abstract", and instead a "Scope" and an "Introduction". I left a short abstract because I _think_ bikeshed can use it in some cases, as metadata to describe what the spec is about.

source-map.bs
nicolo-ribaudo: No! I accidentally copied it from an outdated clone of the repo.

source-map.bs
```diff
@@ -97,19 +102,23 @@ urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
...
+JavaScript) back to the original sources.
```
nicolo-ribaudo, line 117:
```suggestion
This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
```

source-map.bs
nicolo-ribaudo: Oh well, that was already there in the abstract :P

jkup: 🚢 !

source-map.bs
```diff
@@ -97,19 +102,22 @@ urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
...
-specification.  In its current form, it's not a defined standard and is subject to
-modifications.  If you want to get involved you will find more information under
-the following [Specification Repository](https://github.com/tc39/source-map).
+This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
 
-Background {#background}
-========================
+The Source Maps format has the following goals:
+* Support source-level debugging allowing bidirectional mapping
+* Support server-side stack trace deobfuscation
```
szuend, line 120: Should we drop the `server-side`? It's a very loaded term and we don't provide a clear definition what we mean.

szuend (approved):

source-map.bs
nicolo-ribaudo: Here I just copy-pasted this section from further below in the document.

szuend: Ah, I missed that this was a copy-paste. Removing in a separate PR sgtm

PR #127, opened 2024-09-13
Add instructions for generating the spec PDF for Ecma

nicolo-ribaudo:
Technically we should copy-paste the spec in a .docx document that Ecma has, but this alternative process gives a somewhat good result too.

Example: [source-map-draft.pdf](https://github.com/user-attachments/files/16996538/source-map-draft.pdf)

The first commit in this PR is from https://github.com/tc39/source-map/pull/126


source-map.bs
```diff
@@ -21,6 +21,9 @@ Markup Shorthands: markdown yes
 Group: tc39
 </pre>
 
+<!-- Uncomment the next line when generating the PDF version of the spec -->
```
jkup, line 24: 👍🏻

source-map.bs
```diff
@@ -132,6 +135,50 @@ Pavel Podivilov (Google).
...
+International or its successors or assigns.</p>
+
+<p>This document and the information contained herein is provided on an "AS IS" basis and ECMA
+INTERNATIONAL DISCLAIMS ALL WARRANTIES, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO ANY
+WARRANTY THAT THE USE OF THE INFORMATION HEREIN WILL NOT INFRINGE ANY OWNERSHIP RIGHTS OR ANY
+IMPLIED WARRANTIES OF MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.</p>
+</div>
+<div style="display: none" id="copyright-after"></div>
+
+<!-- Use a script tag because bikeshed doesn't like <h1> tags -->
```
jkup, line 177: 😂

jkup (approved):
Looks good!

nicolo-ribaudo:
Note that I will probably follow up with changes after talking with Aki.

PR #128, opened 2024-09-17
Editorial: various edits for style and clarity

takikawa:
This PR shouldn't make any semantic changes and should be purely editorial. It adjusts markup and rewords prose for consistency, clarity, or more spec-like wording. (the commit message has more details)

It builds on PR #126 so should be rebased & merged after that.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -267,8 +267,8 @@ following structure:
     `3` as an integer. The source map may be rejected if the field has any other value.
 * <dfn for="json"><code>file</code></dfn> is an optional name of the generated code
     that this source map is associated with. It's not specified if this can
-    be a URL, relative path name, or just a base name. As such it has a mostly informal
-    character.
+    be a URL, relative path name, or just a base name. Source map generators may
+    choose the appropriate interpretation for their contexts of use.
 * <dfn for="json"><code>sourceRoot</code></dfn> is an optional source root string,
     used for relocating source files on a server or removing repeated values in
     the [=json/sources=] entry.  This value is prepended to the individual entries in the
--- a/source-map.bs
+++ b/source-map.bs
@@ -126,22 +126,22 @@ The document at [https://tc39.es/source-map/](https://tc39.es/source-map/) is th
 up-to-date Source Map specification. It contains the content of the most recently published snapshot
 plus any modifications that will be included in the next snapshot.
 
-If you want to get involved you will find more information under the following [Specification
-Repository](https://github.com/tc39/source-map).
+If you want to get involved you will find more information at the [specification
+repository](https://github.com/tc39/source-map).
 
 <h2 class="no-num">Introduction</h2>
 
 This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
 
-The Source Maps format has the following goals:
+The Source Map format has the following goals:
 * Support source-level debugging allowing bidirectional mapping
 * Support server-side stack trace deobfuscation
 
-The original source map format (v1) was created by Joseph Schorr for use by
+The original Source Map format (v1) was created by Joseph Schorr for use by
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -202,22 +203,22 @@ following structure:
 ```
 
 * <dfn><code>version</code></dfn> is the version field which must always be the number
-    `3` as an integer. The source map may be rejected in case of a value different from `3`.
+    `3` as an integer. The source map may be rejected if the field has any other value.
 * <dfn><code>file</code></dfn> is an optional name of the generated code
-    that this source map is associated with. It's not specified if this can
+    that the source map is associated with. It is not specified if this can
     be a URL, relative path name, or just a base name. As such it has a mostly informal
     character.
```
takikawa, line 210: I didn't adjust it yet because I'm not sure what we want to say, but I feel like the "As such it has a mostly informal character" isn't adding much information here. Should we say something about how generators can choose an appropriate interpretation for their context?

jkup: Yes I prefer your solution to the current one which made little sense to me.

takikawa: I added some text about generators replacing this sentence. It's in a separate commit in case it's too close to non-editorial and there are any objections to changing it now.

jkup:
I think these all look great! Would you mind resolving the conflicts so I can see a cleaner diff? Sorry I merged 2 things in early today that conflict.

takikawa:
> I think these all look great! Would you mind resolving the conflicts so I can see a cleaner diff? Sorry I merged 2 things in early today that conflict.

Thanks for the review! It should be rebased & conflicts resolved now.

source-map.bs
```diff
@@ -126,22 +126,22 @@ The document at [https://tc39.es/source-map/](https://tc39.es/source-map/) is th
...
 
 This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
 
-The Source Maps format has the following goals:
+The Source Map format has the following goals:
 * Support source-level debugging allowing bidirectional mapping
 * Support server-side stack trace deobfuscation
 
-The original source map format (v1) was created by Joseph Schorr for use by
+The original Source Map format (v1) was created by Joseph Schorr for use by
```
jkup, line 140: The casing seems inconsistent. Here we sentence case but then below it's lower case. I'm not sure if we care?

jkup (approved):
This looks great! One tiny comment about casing

source-map.bs
szuend: The way I understood it is it upper cases "Source Map format" but for concrete real instances of "source maps" we lower-case it?

jkup: @nicolo-ribaudo so then in this case, according to your preferred format, we should use "The original source map format"?

nicolo-ribaudo: Yes :)

source-map.bs
```diff
@@ -126,28 +126,28 @@ The document at [https://tc39.es/source-map/](https://tc39.es/source-map/) is th
...
 
-If you want to get involved you will find more information under the following [Specification
-Repository](https://github.com/tc39/source-map).
+If you want to get involved you will find more information at the [specification
+repository](https://github.com/tc39/source-map).
 
 <h2 class="no-num">Introduction</h2>
 
-This Ecma standard defines the Source Map format, used for mapping transpiled source code back to the original sources.
+This Ecma Standard defines the Source Map Format, used for mapping transpiled source code back to the original sources.
```
takikawa, line 134: I made "source map format" generally lowercase now, except in two locations where I think it's really intended as a proper noun as the object that the standard is defining (this spot and one more below).

Also capitalized "Standard" here since it's also done that way below, and in the intro of ES262:

image

I can revise this if we want to make this lowercase too though.

jkup (approved):
Thanks!

PR #130, opened 2024-09-24
Editorial: add a 'decode a list of base64 VLQ' algorithm

szuend:
This PR extracts a 'decode a list of base64 VLQ' algorithm that consumes a string wholly and returns all contained VLQ numbers as a list. We can share this algorithm between decoding `mappings` and `originalScopes`/`generatedRanges` of the scopes proposal.

It also makes the algorithm slightly more readable as we can get rid of `position` and we don't necessarily need the `relative*` variables as they only have a single use-site.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -498,8 +498,8 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
         1. Let |generatedColumn| be 0.
         1. [=For each=] |segment| in |segments|:
             1. [=Decode a list of base64 VLQ=] from |segment| and let |vlqs| be the result.
-            1. If |vlqs|'s [=list/size=] is not 1, 4 or 5, [=optionally report an error=]
-                and continue with the next iteration.
+            1. If |vlqs| is [=list/empty=], [=optionally report an error=] and continue
+                with the next iteration.
             1. Increase |generatedColumn| by |vlqs|[0]. If the result is negative,
                 [=optionally report an error=] and continue with the next iteration.
             1. Let |decodedMapping| be a new [=decoded mapping=] whose
@@ -512,8 +512,8 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
             1. Append |decodedMapping| to |decodedMappings|.
             1. If |vlqs|'s [=list/size=] is 1, continue with the next iteration.
             1. Increase |sourceIndex| by |vlqs[1]|.
-            1. Increase |originalLine| by |vlqs[2]|.
-            1. Increase |originalColumn| by |vlqs[3]|.
+            1. If |vlqs[2]| [=list/exists=], increase |originalLine| by |vlqs[2]|.
+            1. If |vlqs[3]| [=list/exists=], increase |originalColumn| by |vlqs[3]|.
             1. If any of |sourceIndex|, |originalLine|, or |originalColumn| are less than 0, or if
                 |sourceIndex| is greater than or equal to |sources|'s [=list/size=], [=optionally
                 report an error=].
@@ -527,6 +527,7 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
             1. If |nameIndex| is negative or greater than |names|'s [=list/size=],
                 [=optionally report an error=].
             1. Else, set |decodedMapping|'s [=decoded mapping/name=] to |names|[|nameIndex|].
+            1. If |vlqs|'s [=list/size=] is greater then 4, [=optionally report an error=].
     1. Increase |generatedLine| by 1.
 1. Return |decodedMappings|.
 
--- a/source-map.bs
+++ b/source-map.bs
@@ -497,13 +497,10 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
             |groups|[|generatedLine|] on `,`.
         1. Let |generatedColumn| be 0.
         1. [=For each=] |segment| in |segments|:
-            1. Let |position| be a [=position variable=] for |segment|, initially pointing at
-                |segment|'s start.
-            1. [=Decode a base64 VLQ=] from |segment| given |position| and let
-                |relativeGeneratedColumn| be the result.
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -497,13 +497,10 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
...
-            1. Let |position| be a [=position variable=] for |segment|, initially pointing at
-                |segment|'s start.
-            1. [=Decode a base64 VLQ=] from |segment| given |position| and let
-                |relativeGeneratedColumn| be the result.
-            1. If |relativeGeneratedColumn| is null, [=optionally report an error=] and continue
-                with the next iteration.
-            1. Increase |generatedColumn| by |relativeGeneratedColumn|. If the result is negative,
+            1. [=Decode a list of base64 VLQ=] from |segment| and let |vlqs| be the result.
+            1. If |vlqs|'s [=list/size=] is not 1, 4 or 5, [=optionally report an error=]
+                and continue with the next iteration.
```
nicolo-ribaudo, line 502: If the length is, for example, 2, do you know if iterations completely skip the segnemt or do they still add the value of the first VLQ to _generatedColumn_?

source-map.bs
```diff
@@ -497,13 +497,10 @@ sources=] <dfn for="decode source map mappings">|sources|</dfn>, run the followi
...
+            1. Increase |generatedColumn| by |vlqs[0]|. If the result is negative,
```
nicolo-ribaudo, line 503: Nit: 
```suggestion
            1. Increase |generatedColumn| by |vlqs|[0]. If the result is negative,
```

(the |...| syntax should wrap the variable name)

source-map.bs
szuend: Great point, Chrome DevTools [throws](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.ts;l=550;drc=81f47ed35b4c6d4e7ddc5aa538d056b1c37a364e) in the case where `vlqs` has a size of 2. It would add a mapping first, but then aborts decoding mappings via a throw, because it expects being able to read 3 more VLQs at least.

Firefox does [the same](https://github.com/mozilla/source-map/blob/master/wasm-mappings/src/lib.rs#L660) in their wasm implementation.

WebKit inspector is somewhat undefined as they'll start [doing math on `undefined`](https://github.com/WebKit/WebKit/blob/e4c64c9c92bd4bad8d1dfe9615d6b5dae9de00dc/Source/WebInspectorUI/UserInterface/Models/SourceMap.js#L249) in their VLQ decoding code, which I assume will result in some `NaN`s.

VSCode uses `@jrdigewell/sourcemap-codec` under the hood which handles it [similar to WebKit](https://github.com/jridgewell/sourcemap-codec/blob/main/src/sourcemap-codec.ts#L45).

All in all there seems to be a split between either throwing or converting to NaNs. (To be fair I only skimmed all of these so I might have missed validation steps).

From a spec perspective I think it's ok that we reject segments with an invalid amount of VLQ numbers but I don't have a strong opinion on this.

jridgewell: > VSCode uses `@jrdigewell/sourcemap-codec` under the hood which handles it [similar to WebKit](https://github.com/jridgewell/sourcemap-codec/blob/main/src/sourcemap-codec.ts#L45).

Codec'll produce a segment of length 4 filling with 0s. I don't perform any validation, checking wastes time and we can just assume maps are well formed because they're not user input.

And, VSCode uses sourcemap-codec? 😄

szuend: An alternative is to return a `[=queue=]` instead of a `[=list=]` and consume values via `dequeue`. This would also fit better with the iterator style that most implementations employ.

nicolo-ribaudo: What about:
- capturing the length of the list as soon as we get it
- fill the list with zeroes

```
- Let |list| be ...
- Let |count| be |list|'s length
- Append as many 0 as needed at the end of |list|, so that |list|'s |length| is at least 5.
```

szuend: And then compare against `|count|` going forward to correctly handle null mappings or mappings without a name? I think then I'd prefer

```
1. Increase |sourceIndex| by |vlqs[1]|.
1. If |vlqs[2]| [=list/exists=], increase |originalLine| by |vlqs[2]|.
1. If |vlqs[3]| [=list/exists=], increase |originalColumn| by |vlqs[3]|.
```

That would be equivalent to filling with zeroes.

nicolo-ribaudo: This seems good too.

szuend: Ok, I removed the check for length 1, 4 or 5, and replaced it with the conditional increase. I also fixed the formatting for variables (`|..|`). PT(another)L

nicolo-ribaudo (approved):

jridgewell (commented):
Can we keep the optional error for a non-1/4/5 length segment?

My code will be slightly out of spec since we'll eat the `,` as a `0` value and continue reading values from the next segment. I see if I can instead limit my position to `< commaIndex` without adding more checks.

nicolo-ribaudo:
@szuend If you still need this PR for scopes, could you do what Justin suggested above (to avoid introducing observable changes) and rebase?

szuend:
The decoding algorithms for scopes are very unwieldy, and having the format implicitly defined by a decoding algorithm makes it very hard to understand.

I'm wondering if we should go a different route and follow ecma262: Define the encoded format as a grammar and use syntax-directed operations for turning the grammar rules into structs. Ecma262 doesn't spell out a parsing algorithm after all where-as we do that currently in the source map spec.

nicolo-ribaudo:
ECMA-262 doesn't define parsing algorithms, however it has "syntax-directed operations" to convert an "AST node" into an actual value.

For example, when some JS code contains `12.345`, it is parsed implicitly according to the grammar but then it has some explicit recursive algorithms to tell that the token `12.345` represents the number _12.345_: https://tc39.es/ecma262/#sec-numericvalue

We could do something similar.

szuend:
Superseded by #180

PR #131, opened 2024-09-27
Editorial: fixes to suppress bikeshed warnings

takikawa:
This is a minor editorial PR that fixes most of the bikeshed warnings and adjusts linking (in particular, makes sure "map" links to the infra spec where appropriate).

There are just two warnings left after this PR:

```
LINE 683: Unexported dfn that's not referenced locally - did you mean to export it?
<dfn bs-line-number="683" data-dfn-for="index-map" data-dfn-type="dfn" id="index-map-offset" data-lt="offset" data-noexport="by-default" class="dfn-paneled"><code bs-line-number="683" class="highlight">offset</code></dfn>
LINE 687: Unexported dfn that's not referenced locally - did you mean to export it?
<dfn bs-line-number="687" data-dfn-for="index-map" data-dfn-type="dfn" id="index-map-map" data-lt="map" data-noexport="by-default" class="dfn-paneled"><code bs-line-number="687" class="highlight">map</code></dfn>
```

We could remove these dfns but we could also link them when we flesh out the index map sections in the future.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -577,10 +577,10 @@ To <dfn>decode a base64 VLQ</dfn> from a [=string=] |segment| given a [=position
 NOTE: In addition to returning the decoded value, this algorithm updates the [=position variable=]
 in the calling algorithm.
 
-### [=json/Mappings=] for generated JavaScript code ### {#mapping-javascript}
+### [=decoded mapping|Mappings=] for generated JavaScript code ### {#mappings-javascript}
 
-Generated code positions that may have [=json/Mapping=] entries are defined in terms of *input elements*
-as per [=ECMAScript Lexical Grammar=]. [=json/Mapping=] entries must point to either:
+Generated code positions that may have [=decoded mapping|mapping=] entries are defined in terms of *input elements*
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -577,10 +577,10 @@ To <dfn>decode a base64 VLQ</dfn> from a [=string=] |segment| given a [=position
 NOTE: In addition to returning the decoded value, this algorithm updates the [=position variable=]
 in the calling algorithm.
 
-### [=Mappings=] for generated JavaScript code
+### [=json/Mappings=] for generated JavaScript code ### {#mapping-javascript}
```
szuend, line 580: Shouldn't these point to the "decoded mappings struct" because we talk about "mapping entries"? In the "names" PR I changed these to `[=decoded mapping|Mapping=]`.

szuend (approved):

jkup (approved):

source-map.bs
takikawa: Thanks, that sounds good to me. I revised this PR to match the names PR (#122) so there's no conflict here.

PR #132, opened 2024-09-27
Add a decoding algorithm for index maps

takikawa:
[**Rendered draft**](https://takikawa.github.io/source-map-spec/#decoding-an-index-map)

This PR adds a decoding algorithm for index maps along the lines of the existing algorithms for normal source maps.

I'll leave a few comments for cases in the spec where it's not clear what we should do:
  * Error cases for offsets
  * The check for index map overlap

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -832,7 +832,7 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
     1. If |section|[`"map"`] does not [=map/exist=], then report an error.
     1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
     1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
-    1. Set |sourceMap|'s [=decoded source map/sources=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
+    1. [=list/Extend=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
     1. Let |offsetMappings| be a new [=list=].
     1. Let |firstLine| be null.
     1. [=For each=] |mapping| of |decodedSection|'s [=decoded source map/mappings=]:
@@ -846,7 +846,7 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
         1. Set |offsetMapping|'s [=decoded mapping/originalColumn=] to |mapping|'s [=decoded mapping/originalColumn=].
         1. Set |offsetMapping|'s [=decoded mapping/name=] to |mapping|'s [=decoded mapping/name=].
         1. [=list/Append=] |offsetMapping| to |offsetMappings|.
-    1. Set |sourceMap|'s [=decoded source map/mappings=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
+    1. [=list/Extend=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
     1. Set |previousOffset| to |section|[`"offset"`].
     1. Let |sortedMappings| be the result of [=list/sorting=] |offsetMappings| in ascending order, with an item |a| being less than |b| if |a| is [=generated position less than=] |b|.
     1. Set |previousLastMapping| to the last item of |sortedMappings| if |sortedMappings| [=is not empty=] and to null otherwise.
--- a/source-map.bs
+++ b/source-map.bs
@@ -808,6 +808,7 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
 1. Let |sourceMap| be a new [=decoded source map=].
 1. Set |sourceMap|'s [=decoded source map/file=] to [=optionally get a string=] `"file"` from |jsonMap|.
 1. Let |previousOffset| be null.
+1. Let |previousFirstMapping| be null.
 1. Let |previousLastMapping| be null.
 1. [=For each=] |section| of |jsonMap|[`"sections"`]:
     1. If |section| is not a [=/map=], [=optionally report an error=] and continue with the next iteration.
@@ -817,18 +818,8 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
     1. If |offsetLine| is not a number, [=optionally report an error=].
     1. If |offsetColumn| is not a number, [=optionally report an error=].
     1. If |previousOffset| is not null,
-        1. If |previousLastMapping| is not null,
-            1. If |offsetLine| is less than |previousLastMapping|'s [=decoded mapping/generatedLine=], [=optionally report an error=].
-            1. If |offsetLine| is equal to |previousLastMapping|'s [=decoded mapping/generatedLine=] and |offsetColumn| is less than |previousLastMapping|'s [=decoded mapping/generatedColumn=], [=optionally report an error=].
-        1. Otherwise,
-            1. If |offsetLine| is less than |previousOffset|[`"line"`], [=optionally report an error=].
-            1. If |offsetLine| is equal to |previousOffset|[`"line"`] and |offsetColumn| is less than |previousOffset|[`"column"`], [=optionally report an error=].
-
-        Note: This part of the decoding algorithm checks that index source map sections do not overlap. While it is
-              expected that generators should not produce index source maps with overlapping sections, source map
-              consumers may choose instead to only check the simpler condition that the sections are ordered and therefore
-              that offsets appear in increasing order.
-
+        1. If |offsetLine| is less than |previousOffset|[`"line"`], [=optionally report an error=].
+        1. If |offsetLine| is equal to |previousOffset|[`"line"`] and |offsetColumn| is less than |previousOffset|[`"column"`], [=optionally report an error=].
     1. If |section|[`"map"`] does not [=map/exist=], then report an error.
     1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
     1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
@@ -844,13 +835,20 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
         1. Set |offsetMapping|'s [=decoded mapping/originalColumn=] to |mapping|'s [=decoded mapping/originalColumn=].
         1. Set |offsetMapping|'s [=decoded mapping/name=] to |mapping|'s [=decoded mapping/name=].
         1. [=list/Append=] |offsetMapping| to |offsetMappings|.
+
+        Note: Implementations may choose to represent index source map sections without appending the mappings together,
+              for example, by storing each section separately and conducting a binary search.
     1. [=list/Extend=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
     1. Set |previousOffset| to |section|[`"offset"`].
     1. Let |sortedMappings| be the result of [=list/sorting=] |offsetMappings| in ascending order, with an item |a| being less than |b| if |a| is [=generated position less than=] |b|.
-    1. Set |previousLastMapping| to the last item of |sortedMappings| if |sortedMappings| [=is not empty=] and to null otherwise.
+    1. If |previousFirstMapping| and |previousLastMapping| are not null,
+        1. If |previousLastMapping| is not [=generated position less than=] the first [=decoded mapping=] of |sortedMappings|, [=optionally report an error=].
 
-    Note: See the above note about section overlap and order. These last two steps of storing the last mapping of each
-          section are only necessary to enforce the conditions concerning section overlap.
+        Note: This part of the decoding algorithm checks that index source map sections are ordered and do not overlap.
+              While it is expected that generators should not produce index source maps with overlapping sections, source map
+              consumers may, for example, only check the simpler condition that the section offsets are ordered.
+    1. Set |previousFirstMapping| to the first item of |sortedMappings| if |sortedMappings| [=is not empty=].
+    1. Set |previousLastMapping| to the last item of |sortedMappings| if |sortedMappings| [=is not empty=].
 
 1. Return |sourceMap|.
 
--- a/source-map.bs
+++ b/source-map.bs
@@ -42,6 +42,7 @@ spec:url; type:dfn; for:/; text:url
 spec:infra; type:dfn;
     text:list
     for:list; text:for each
+    for:list; text:is not empty
 </pre>
 <pre class="anchors">
 urlPrefix:https://tc39.es/ecma262/#; type:dfn; spec:ECMA-262
@@ -363,7 +364,8 @@ following steps:
 1. Let |jsonMap| be the result of [=parse a JSON string to an Infra value|parsing a JSON string to
     an Infra value=] |str|.
 1. If |jsonMap| is not a [=/map=], report an error and abort these steps.
-1. [=Decode a source map=] given |jsonMap| and |baseURL|, and return its result if any.
+1. If |jsonMap|[`"sections"`] [=map/exists=], [=decode an index source map=] given |jsonMap| and |baseURL|, and return its result if any.
+1. Otherwise, [=decode a source map=] given |jsonMap| and |baseURL|, and return its result if any.
 
 To <dfn export>decode a source map</dfn> given a [=string=]-keyed [=/map=] |jsonMap| and a [=/URL=]
 |baseURL|, run the following steps:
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -674,22 +676,84 @@ The index map follows the form of the standard map.  Like the regular source map
...
+1. Set |sourceMap|'s [=decoded source map/file=] to [=optionally get a string=] `"file"` from |jsonMap|.
+1. Let |previousOffset| be null.
+1. Let |previousLastMapping| be null.
+1. [=For each=] |section| of |jsonMap|[`"sections"`]:
+    1. If |section| is not a [=/map=], [=optionally report an error=] and continue with the next iteration.
+    1. If |section|[`"offset"`] does not [=map/exist=], then report an error.
+    1. Let |offsetLine| be |section|[`"offset"`][`"line"`].
+    1. Let |offsetColumn| be |section|[`"offset"`][`"column"`].
+    1. If |offsetLine| is not a number, [=optionally report an error=].
+    1. If |offsetColumn| is not a number, [=optionally report an error=].
```
takikawa, line 713: I left these checks a bit vague, but it's not very meaningful for line and column to be non-integral values. Should we coerce these to integers and then add them? Or reject non-integral values? (as in `Number.isInteger()`)

source-map.bs
```diff
@@ -674,22 +676,84 @@ The index map follows the form of the standard map.  Like the regular source map
...
+    1. If |previousOffset| is not null,
+        1. If |previousLastMapping| is not null,
+            1. If |offsetLine| is less than |previousLastMapping|'s [=decoded mapping/generatedLine=], [=optionally report an error=].
```
takikawa, line 716: Should these range checks could be "less than or equal to" to be strict inequality rather than allowing equality?

source-map.bs
```diff
@@ -674,22 +676,84 @@ The index map follows the form of the standard map.  Like the regular source map
...
+            1. If |offsetLine| is equal to |previousLastMapping|'s [=decoded mapping/generatedLine=] and |offsetColumn| is less than |previousLastMapping|'s [=decoded mapping/generatedColumn=], [=optionally report an error=].
+        1. Otherwise,
+            1. If |offsetLine| is less than |previousOffset|[`"line"`], [=optionally report an error=].
+            1. If |offsetLine| is equal to |previousOffset|[`"line"`] and |offsetColumn| is less than |previousOffset|[`"column"`], [=optionally report an error=].
+
+        Note: This part of the decoding algorithm checks that index source map sections do not overlap. While it is
+              expected that generators should not produce index source maps with overlapping sections, source map
+              consumers may choose instead to only check the simpler condition that the sections are ordered and therefore
+              that offsets appear in increasing order.
```
takikawa, line 725: As the note says, I'm not sure anyone will implement the overlap check here (which ensures that the offset position comes after the last ordered mapping, or the last offset if no mappings exist, of the last section) but it encodes what we expect of generators at least.

Note: none of the browsers implement an overlap check as far as I know. I think only Firefox dev tools even checks order of offsets.

source-map.bs
```diff
@@ -674,22 +676,84 @@ The index map follows the form of the standard map.  Like the regular source map
...
+
+    1. If |section|[`"map"`] does not [=map/exist=], then report an error.
+    1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
+    1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
+    1. Set |sourceMap|'s [=decoded source map/sources=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
+    1. Let |offsetMappings| be a new [=list=].
+    1. [=For each=] |mapping| of |decodedSection|'s [=decoded source map/mappings=]:
+        1. Let |offsetMapping| be a new [=decoded mapping=].
+        1. Set |offsetMapping|'s [=decoded mapping/generatedLine=] to |mapping|'s [=decoded mapping/generatedLine=] + |offsetLine|.
+        1. Set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
```
jkrems, line 735: This looks like it also adds the `offsetColumn` to all subsequent lines? Should this be guarded by "if this is the first generated line of the mapping"?

takikawa: Oh thanks for catching this, yeah this should only apply to the first line.

(alternatively, this part of the algorithm could be changed to pass the offsets to the "decode a source map" as initial offset arguments which are normally 0)

source-map.bs
```diff
@@ -674,22 +676,87 @@ The index map follows the form of the standard map.  Like the regular source map
...
+    1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
+    1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
+    1. Set |sourceMap|'s [=decoded source map/sources=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
+    1. Let |offsetMappings| be a new [=list=].
+    1. Let |firstLine| be null.
+    1. [=For each=] |mapping| of |decodedSection|'s [=decoded source map/mappings=]:
+        1. Let |offsetMapping| be a new [=decoded mapping=].
+        1. Set |offsetMapping|'s [=decoded mapping/generatedLine=] to |mapping|'s [=decoded mapping/generatedLine=] + |offsetLine|.
+        1. If |firstLine| is null, set |firstLine| to |offsetMapping|'s [=decoded mapping/generatedLine=].
+        1. If |offsetMapping|'s [=decoded mapping/generatedLine=] is equal to |firstLine|, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
```
jridgewell, line 737: This should just check for `mapping.generatedLine === 1`. If the first mapping is on generated line 3, that doesn't get column added.

source-map.bs
```diff
@@ -674,22 +676,87 @@ The index map follows the form of the standard map.  Like the regular source map
...
+
+        Note: This part of the decoding algorithm checks that index source map sections do not overlap. While it is
+              expected that generators should not produce index source maps with overlapping sections, source map
+              consumers may choose instead to only check the simpler condition that the sections are ordered and therefore
+              that offsets appear in increasing order.
+
+    1. If |section|[`"map"`] does not [=map/exist=], then report an error.
+    1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
+    1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
+    1. Set |sourceMap|'s [=decoded source map/sources=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
```
jridgewell, line 730: This is a simple append, but the `mapping.originalSource`s are just strings, so we could treat as a set.

source-map.bs
```diff
@@ -674,22 +676,87 @@ The index map follows the form of the standard map.  Like the regular source map
...
+        1. Otherwise, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=].
+        1. Set |offsetMapping|'s [=decoded mapping/originalSource=] to |mapping|'s [=decoded mapping/originalSource=].
+        1. Set |offsetMapping|'s [=decoded mapping/originalLine=] to |mapping|'s [=decoded mapping/originalLine=].
+        1. Set |offsetMapping|'s [=decoded mapping/originalColumn=] to |mapping|'s [=decoded mapping/originalColumn=].
+        1. Set |offsetMapping|'s [=decoded mapping/name=] to |mapping|'s [=decoded mapping/name=].
+        1. [=list/Append=] |offsetMapping| to |offsetMappings|.
+    1. Set |sourceMap|'s [=decoded source map/mappings=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
+    1. Set |previousOffset| to |section|[`"offset"`].
+    1. Let |sortedMappings| be the result of [=list/sorting=] |offsetMappings| in ascending order, with an item |a| being less than |b| if |a| is [=generated position less than=] |b|.
```
jridgewell, line 746: Let's do this before the append? They're N distinct regions, no reason to let the sorting algorithm do position guesses in another sections mappings.

source-map.bs
takikawa: I've updated it to use set union now, and also did the same with decoded sources.

source-map.bs
takikawa: The reason it's after the append currently is that this `sortedMappings` is only used to check the section overlap constraint, and the appended mappings don't need to actually be sorted. IIUC you were suggesting sorting the mappings too to avoid the extra computation right?

BTW: I'm actually thinking of proposing to remove the "non-overlapping" constraint or weaken how we define it, because this check seems overly complex. Will bring it up at the next meeting.

source-map.bs
```diff
@@ -881,18 +881,19 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
...
     1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
-    1. Set |sourceMap|'s [=decoded source map/sources=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/sources=] with |decodedSection|'s [=decoded source map/sources=].
+    1. Set |sourceMap|'s [=decoded source map/sources=] to the [=set/union=] of |sourceMap|'s [=decoded source map/sources=] and |decodedSection|'s [=decoded source map/sources=].
     1. Let |offsetMappings| be a new [=list=].
     1. [=For each=] |mapping| of |decodedSection|'s [=decoded source map/mappings=]:
         1. Let |offsetMapping| be a new [=decoded mapping=].
         1. Set |offsetMapping|'s [=decoded mapping/generatedLine=] to |mapping|'s [=decoded mapping/generatedLine=] + |offsetLine|.
-        1. Set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
+        1. If |offsetMapping|'s [=decoded mapping/generatedLine=] is equal to 1, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
```
jridgewell, line 889: My mistake, this should be 0.
```suggestion
        1. If |offsetMapping|'s [=decoded mapping/generatedLine=] is equal to 0, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
```

source-map.bs
```diff
@@ -881,18 +881,19 @@ To <dfn export>decode an index source map</dfn> given a [=string=]-keyed [=/map=
...
+        1. Otherwise, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=].
         1. Set |offsetMapping|'s [=decoded mapping/originalSource=] to |mapping|'s [=decoded mapping/originalSource=].
         1. Set |offsetMapping|'s [=decoded mapping/originalLine=] to |mapping|'s [=decoded mapping/originalLine=].
         1. Set |offsetMapping|'s [=decoded mapping/originalColumn=] to |mapping|'s [=decoded mapping/originalColumn=].
         1. Set |offsetMapping|'s [=decoded mapping/name=] to |mapping|'s [=decoded mapping/name=].
         1. [=list/Append=] |offsetMapping| to |offsetMappings|.
-    1. Set |sourceMap|'s [=decoded source map/mappings=] to the result of [=list/extending=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
+    1. Set |sourceMap|'s [=decoded source map/mappings=] to the [=set/union=] of |sourceMap|'s [=decoded source map/mappings=] and |offsetMappings|.
```
jridgewell, line 896: This should be regular list extend, right? We know that the offsets are increasing, so there can't be a duplicate (except within the same map, which we shouldn't be pruning out).

source-map.bs
jridgewell: > The reason it's after the append currently is that this sortedMappings is only used to check the section overlap constraint, and the appended mappings don't need to actually be sorted

Right, which means we only need to sort the appended mappings. We know they don't overlap, everything before must be sorted, so we only need to sort these new mappings before we append.

Right now, there's a bug where you sort a list after it's appended to another list, which means the sort operation is discarded. We either sort the full list after appending, or we sort the smaller list before appending.

> IIUC you were suggesting sorting the mappings too to avoid the extra computation right?

Hopefully mappings is never unsorted, and one of the things I think we should change in the future.

> I'm actually thinking of proposing to remove the "non-overlapping" constraint or weaken how we define it, because this check seems overly complex

The constraint is important for mine, Mozilla's, and Sentry's impls. When doing lookups, we just binary search the offsets, then binary search the section's map. Weakening this means we have to search a union, which can be much slower.

source-map.bs
takikawa: We talked about this a little bit in the TG4 meeting and I'll draft some text to optionally throw here if it's not an integral value (and also check what implementations do if anything).

source-map.bs
takikawa: > Right, which means we only need to sort the appended mappings. We know they don't overlap, everything before must be sorted, so we only need to sort these new mappings before we append.
>
> Right now, there's a bug where you sort a list after it's appended to another list, which means the sort operation is discarded. We either sort the full list after appending, or we sort the smaller list before appending.

I think I'm not quite following. Right now the draft algorithm sorts only the `sortedMappings` (which is a new copy because list sort is functional, not in-place) and the original unsorted `offsetMappings` is appended. So the appended mappings should be in whatever order they came from decoding, which could be unordered.

> The constraint is important for mine, Mozilla's, and Sentry's impls. When doing lookups, we just binary search the offsets, then binary search the section's map. Weakening this means we have to search a union, which can be much slower.

This makes sense, and I'm convinced the ordering constraint should be kept in. I've simplified/clarified it in the latest push to use the previous `{first_0, last_0}` mappings and the current `{first_1, last_1}` mapping as ranges and do the standard `last_0 < first_1` check.

Do you also rely on the mappings being in the range of {currentOffset, nextOffset}? (that is, the next offset must be larger than the last mapping of the current range)

jridgewell: The spec is out of order:
```
    1. [=list/Extend=] |sourceMap|'s [=decoded source map/mappings=] with |offsetMappings|.
    1. Set |previousOffset| to |section|[`"offset"`].
    1. Let |sortedMappings| be the result of [=list/sorting=] |offsetMappings| in ascending order, with an item |a| being less than |b| if |a| is [=generated position less than=] |b|.
```

It concats mappings with `offsetMappings`, then sorts `offsetMappings`, and then we throw away `sortedMappings`.

We can either sort then append, or we can append then sort. In the current ordering, I think you intend to append then sort, but you're sorting the wrong list. If we switch the ordering, then it solves both the bug and my suggestion, we'll only sort the new mappings.

takikawa: It's actually intentional that the sorted list is thrown away, as it's only used to determine the first and last items of each sub-map's mappings range. These first and last items are, in turn, only used to check the overlap requirements.

It's not intended to sort the actual mappings because they're not sorted in the ordinary, non-index map algorithm either and it should probably be consistent between the decoding algorithms.

jridgewell (approved):

nicolo-ribaudo:
Could you rebase? :)

takikawa:
I rebased this and made some adjustments. It now checks that mapping ranges are always contained within the section offsets too, which I think is required for the assumptions made in implementations about mappings appearing in a unique bucket.

That means mappings are ordered like this: `previousOffset --- firstMapping --- lastMapping --- nextOffset` and so on.

takikawa:
@jridgewell Just a friendly ping for a last pass on this, thanks!

source-map.bs
```diff
@@ -787,16 +790,75 @@ the file format is JSON with a top-level object.  It shares the [=json/version=]
...
+              consumers may, for example, only check the simpler condition that the section offsets are ordered.
+    1. If |section|[`"map"`] does not [=map/exist=], then report an error.
+    1. Let |decodedSection| be the result of [=decoding a source map=] given |section|[`"map"`] and |baseURL|.
+    1. If the previous step reported an error, [=optionally report an error=] and continue with the next iteration.
+    1. Set |sourceMap|'s [=decoded source map/sources=] to the [=set/union=] of |sourceMap|'s [=decoded source map/sources=] and |decodedSection|'s [=decoded source map/sources=].
+    1. Let |offsetMappings| be a new [=list=].
+    1. [=For each=] |mapping| of |decodedSection|'s [=decoded source map/mappings=]:
+        1. Let |offsetMapping| be a new [=decoded mapping=].
+        1. Set |offsetMapping|'s [=decoded mapping/generatedLine=] to |mapping|'s [=decoded mapping/generatedLine=] + |offsetLine|.
+        1. If |offsetMapping|'s [=decoded mapping/generatedLine=] is equal to 0, set |offsetMapping|'s [=decoded mapping/generatedColumn=] to |mapping|'s [=decoded mapping/generatedColumn=] + |offsetColumn|.
```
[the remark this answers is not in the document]
takikawa: Thanks for catching that, I'll include a change based on this. I think `decodedSection` should be `mapping` though.

PR #135, opened 2024-10-06
Move clear normative algorithms out of "Conventions"

nicolo-ribaudo:
I was waiting for https://github.com/tc39/source-map/pull/122 to be merged before opening this PR, to do a last editorial pass after merging all the normative contents we wanted for the first edition, but given that the presentation is in two days I prefer to get this PR merged as soon as possible and cut a branch.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -690,17 +690,8 @@ Section objects have the following fields:
 The sections must be sorted by starting position and the represented sections
 must not overlap.
 
-Conventions {#conventions}
-==========================
-
-The following conventions should be followed when working with source maps or
-when generating them.
-
-Source Map Naming {#source-map-naming}
---------------------------------------
-
-Commonly, a source map will have the same name as the generated file but with a `.map`
-extension.  For example, for `page.js` a source map named `page.js.map` would be generated. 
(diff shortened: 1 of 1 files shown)


source-map.bs
```diff
@@ -690,17 +690,8 @@ Section objects have the following fields:
...
-Source Map Naming {#source-map-naming}
---------------------------------------
-
-Commonly, a source map will have the same name as the generated file but with a `.map`
-extension.  For example, for `page.js` a source map named `page.js.map` would be generated. 
+Retrieving Source Maps {#linking-and-fetching}
+========================================================
 
 Linking generated code to source maps {#linking-generated-code}
 ---------------------------------------------------------------
```
takikawa, line 697: In this moved section a few lines down it says "it is useful to have some conventions for the expected use-case of web server-hosted JavaScript" where I think "conventions" is referring to the old section it was in. Would it make sense to remove "conventions" after moving this, and replace it with more normative text?

takikawa (approved):
LGTM modulo the comment I left

source-map.bs
nicolo-ribaudo: Done!

PR #136, opened 2024-10-06
Relicense to Ecma

nicolo-ribaudo:
This PR changes the license of the specification to match Ecma's requirements. We can do it because:
1. all contributors since when we created TC39-TG4 signed Ecma's IPR form
2. previous contributions came from employees of Ecma members

For (2), there is a 60 days period during which Ecma members can say "actually, we are not ok with this". If that happens (hopefully not!), we will need to ask to the Ecma General Assembly for an exception to be able to publish under the old license.

cc @littledan @gesa @SaminaHusain


jkup (approved):

nicolo-ribaudo:
I talked with Aki -- let's present this as in progress to the committee and wait until I get written confirmation from Google&Firefox.

nicolo-ribaudo (commented):
This was approved by Google and Mozilla. The license/copyright text is now the same as in the ECMA-262 spec.

szuend:
The README.md still references the CC license.

PR #142, opened 2024-10-29
Debug IDs: Clarify how Debug IDs should be generated

lforst:
Based on: https://github.com/tc39/source-map/pull/141

--- 

- Clarify how Debug IDs should be generated and explain proposed requirement of 128 bit UUID.
- Clarify that Debug IDs do not need to be deterministic and why they don't need to be deterministic.

### DIFF
--- a/proposals/debug-id.md
+++ b/proposals/debug-id.md
@@ -56,17 +56,24 @@ In the context of this document:
 
 ## Debug IDs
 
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Given that we talked about this in the meeting and that it's an early stage proposal, I'm merging without further approvals.

nicolo-ribaudo (approved):

PR #144, opened 2024-10-29
Debug IDs: Only allow Debug IDs in the canonical UUID format

lforst:
Builds on https://github.com/tc39/source-map/pull/141

---

This proposal change may be a bit controversial.

I am suggesting to limit Debug IDs within Source Maps and generated files to be in a specific format - specifically the canonical UUID format (e.g. `85314830-023f-4cf1-a267-535f4e37bb17`)

This is for the very simple reason of ensuring compatibility, consistency and simplicity in the ecosystem.

Drawbacks are:
- Reduced flexibility
- 4 more bytes per file

Would be great to have some feedback and discussions around this!

### DIFF
--- a/proposals/debug-id.md
+++ b/proposals/debug-id.md
@@ -57,10 +57,8 @@ In the context of this document:
 ## Debug IDs
 
 Debug IDs are globally unique identifiers for build artifacts.
-They are specified to be UUIDs.
-In the context of this proposal, they are represented in hexadecimal characters.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Could you rebase? :(

nicolo-ribaudo (approved):
Given that we talked about this in the meeting and nobody raised concerns, and that it's an early stage proposal, I'm merging without further approvals.

PR #151, opened 2024-11-13
Editorial: Extract 'generatedPosition' and 'originalPosition' structs

szuend:
This PR introduces two new structs `generated position` and `original position` and uses them in the `decoded mapping` struct instead of inlined line/column/sources. This has a couple of advantages:

- We make it explicit with the "types" that for original position in a mapping entry either all (source, line, column) are null or a non-negative integer. Otherwise it's only implicitly defined by the decoding algorithm.
- We can re-use both in the definition for "original scopes" (start, end) and "generated ranges" (start, end callsite).
- We can re-use both in various algorithms we'll declare (both for scopes and how to apply mappings).

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -501,7 +501,7 @@ The [=json/mappings=] data is broken down as follows:
 
 The fields in each segment are:
 
-1. The zero-based starting [=column=] of the line in the generated code that the segment represents.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (approved):

szuend:
We probably want something similar in the ecmarkup version later on.

PR #154, opened 2024-11-16
DRAFT: 2024 pdf

gesa:


nicolo-ribaudo:
@gesa I guess we should close this?

gesa:
@nicolo-ribaudo Yup. I never planned for it to be merged, just available as reference

PR #156, opened 2024-11-20
Editorial: Add CSS to adjust list bullets to match ECMA

takikawa:
Right now the draft PDF spec (in ECMA format) and the web spec don't align on the numbered bullet styles in algorithms. This PR changes the web version to match using CSS.

This also changes numbered bullets that aren't in algorithms, such as those in "5.1.2. Names for generated JavaScript code" (which is not in the 2024 branch) because we don't have a CSS class to distinguish them. I think in that particular case, we could just use unordered lists instead if we don't like the alternating bullets.

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -145,6 +145,17 @@ urlPrefix:https://webassembly.github.io/spec/core/; type:dfn; spec:wasm
 }
 </pre>
 
+<style>
+  /* match list styles of ecmarkup */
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
> we could just use unordered lists instead if we don't like the alternating bullets.

Yeah, it seems reasonable to use an unordered lists for things that don't have a meaningful order of any type.

nicolo-ribaudo (approved):

PR #160, opened 2024-11-25
Make "column" definition a partial sentence

nicolo-ribaudo:
More feedback coming from Ecma. All the other definitions are given as partial sentences, so "column" should follow.


nicolo-ribaudo (commented):
This has been already merged in the PDF version sent to Ecma

PR #161, opened 2024-11-27
Advance Debug ID proposal to stage 2

lforst:
In the last TG4 monthly meeting we decided to advance the Debug ID proposal to stage 2 given the proposal is written out in detail and also has some adoption.

I am happy to wait a few more weeks to give people the opportunity to object.


szuend (approved):

nicolo-ribaudo:
Merging, as discussed in the meeting ~3 weeks ago.

PR #162, opened 2024-12-04
Add aknowledgment for editorial changes in first Ecma edition

nicolo-ribaudo:
This was suggested on Matrix by other delegates with more Ecma experience than us, highlighting that the work Jon and I have been doing in the past months to get the spec to a good Ecma state is similar to an editor's role, even without being officially editors.

I copied the way of listing editor names from ECMA-262 and ECMA-402 (as "signatures" at the end of the introduction).

### DIFF
--- a/source-map.bs
+++ b/source-map.bs
@@ -177,6 +177,11 @@ Pavel Podivilov (Google).
 The source map format does not have version numbers anymore, and it is instead
 hard-coded to always be "3".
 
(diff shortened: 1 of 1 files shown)


littledan (approved):

nicolo-ribaudo:
Oh the repo rename broke the IPR script, I'll fix that tomorrow.

PR #169, opened 2025-01-14
Range Mappings: Allow multi-line ranges

robpalme:
This update changes the semantics of the Range Mapping such that it applies to character sequences that include newlines.

The encoding format/syntax is unchanged. This is a purely semantic change.

It is motivated by cases where most or all of a file contents could be expressed via a single range mapping, e.g. to concisely produce an identity range mapping.

Source:  https://issues.chromium.org/issues/364917746


proposals/range-mappings.md
```diff
@@ -82,17 +85,17 @@ Generate Line 1 Column 0 -> Original Line 2 Column 2 (range mapping)
...
+|       |     + maps to Original Line 3 Column 0
 |       + maps to Original Line 2 Column 10
 + maps to Original Line 2 Column 2
 ```
 
 ### Encoding
 
 To avoid a breaking change to the `mappings` field, a new field named `rangeMappings` is added.
-It contains encoded data per line in the generated code.
+It contains encoded data per-line in the generated code.
```
nicolo-ribaudo, line 98: We don't need to update this because the list of mappings is still per-line, right?

e.g. if we have a mapping at `0:0->0:0` and one at `2:0->3:0` (so on two different lines), to mark them as range mappings we would encode them as `A;;A` in `rangeMappings`, but the first one would span two lines because it spans up to the end of the mapping after?

robpalme: Correct.

This specific text change is purely a typo fix.  The encoding for `mappings` and `rangeMappings` remains per-line.

The encoding remains per-line and per-mapping.  So in your example, if you have two range mappings, you still need two entries in `rangeMappings`.

So I'm agreeing with you.  The extent of a range mapping is from the tagged mapping until the next mapping - regardless of whether that mapping is tagged as range mapping or not.

PR #171, opened 2025-02-04
Scopes: Update encoding to 'tags' approach

szuend:
This PR changes the encoding of the "Scopes" proposal to the "Tag" approach presented in the TG4 meeting of 2024-01-30 ([see slides](https://docs.google.com/presentation/d/1KPGIp_Kr3iSjZo_NOE5Wgwo7PhosrBC6viYd9daBPqM/edit?usp=sharing).

The information contained is exactly the same, we just use a different encoding scheme to be future-proof.

The PR includes some changes w.r.t. to the slides. These are of course up for discussion:

* The PR stores binding expressions in a separate JSON field `"expressions"` rather than piggy-backing on `"names"`. This was discussed in some forums. I'm not 100% sure about this as some strings may end up twice in the source map this way. Especially if no variable renaming happened.
* The slides didn't contain a way how to connect original scope trees to source file indices. The PR includes a new (optional) `original_scope_source_index` item that can be used to connect a root original scope tree with a specific source file. A second way is to include exactly as many root original scope trees as there are source files, so the connection is implicit via indices. The generator can choose the approach.
* The PR moves `FLAGS` in `original_scope_start` to the beginning to. be more consistent with `generated_range_start`.
* `DEFINITION` in `generated_range_start` is not an item index, but rather the nth `original_scope_start` item to further reduce the magnitude of the value.
* The PR splits bindings into 2 separate items: The general case is a simple list of expressions that map 1:1 to the variables list. I.e. the variables use the same expression throughout the full generated range. The subrange mechanism was split into its own item: It contains a variable index, so it can be applied to some variables. This allows us to encode expressions relative to each other and we no longer have the shenanigans with -1 meaning unavailable and < -1 means "number of subranges".

The wording is still ambiguous in some cases, but should be (mostly) enough for the proposal doc.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -196,106 +196,192 @@ interface OriginalPosition {
 
 ### Encoding
 
-We introduce two new fields "originalScopes" and "generatedRanges" respectively:
+We introduce two new fields "scopes" and "expressions" to the source map JSON:
 
-  * "originalScopes" is an array of original scope tree descriptors (a string). Each element in the array describes the scope tree of the corresponding "sources" entry.
-  * "generatedRanges" is a single generated range tree descriptor (a string) of the generated file.
+  * "scopes" is a string. It contains a list of comma-separated items. Each item is prefixed with a unique "tag". The items themselves build a tree structure that describe "original scope" and "generated range" trees.
+  * "expressions" is an array of strings. It is similar to "names" and serves as a string table for JavaScript expressions. `null` is a valid value in this string array and is used to signify that variables are unavailable.
 
-Like the "mappings" field, the data in a "generated range tree descriptor" is grouped by line and lines are separated by `;`. Within a line, items are separated by `,`. A "original scope tree descriptor" is NOT grouped by line, but items are separated by `,`.
+The format of "scopes" is presented in an EBNF-like grammar, with:
 
-There are two different kinds of items that will appear in a "original scope descriptor": "Start Original Scope" and "End Original Scope".
-There are two different kinds of items that will appear in the "generated range descriptor": "Start Generated Range" and "End Generated Range".
+  * Three terminals: Signed, unsigned VLQ and comma '`,`'. VLQ terminals are labelled and we denote them with UPPERCASE.
+    We prefix the terminal with `u` or `s` to signify an unsigned or signed VLQ respectively. E.g. the terminal `uLINE` signifies
+    an unsigned VLQ labelled `LINE`.
+  
+  * Non-terminals are denoted with `snake_case`.
 
-The kind of an item can be determined by looking at how many VLQ-encoded numbers it contains: "End" items contain one number, "Start" items contain two or more numbers.
+  * `symbol?` means zero or one `symbol`.
+  * `symbol+` means one or more `symbol`.
 
-Note: Each DATA represents one VLQ number.
+The start symbol is `scopes`:
 
-#### Start Original Scope
+```
+scopes := item_list
 
-* DATA line in the original code
-  * Note: this is the point in the original code where the scope starts. `line` is relative to the `line` of the preceding "start/end original scope" item.
-* DATA column in the original code
-  * Note: Column is always absolute.
-* DATA flags
-  * Note: binary flags that specify if a field is used for this scope.
-  * Note: Unknown flags would skip the whole scope.
+item_list :=
+    top_level_item
+  | top_level_item ',' item_list
+
+top_level_item :=
+    original_scope_tree
+  | generated_range_tree
+```
+
+The recommendation is to use one top-level `original_scope_tree` per `sources` file, but multiple are technically allowed.
+Multiple top-level `generated_range_tree`s are also allowed, this is especially useful when multiple bundles are straight-up
+concatenated.
+
+
+#### Original Scope Trees
+
+```
+original_scope_tree :=
+  original_scope_start
+  original_scope_variables?
+  original_scope_source_index?
+  original_scope_tree?
+  original_scope_end
+```
+
+A scope is delineated by a `original_scope_start` and a `original_scope_end` item. The `original_scope_variables` and `original_scope_source_index`
+item always describe the immediately "surrounding" start/end pair.
+
+```
+original_scope_start :=
+  'B'     // Tag: 0x1 unsigned
+  uFLAGS
+  uLINE
+  uCOLUMN
+  sNAME?  // Present if FLAGS<0> is set.
+  sKIND?  // Present if FLAGS<1> is set.
+
+original_scope_variables :=
+  'D'        // Tag: 0x3 unsigned
+  sVARIABLE+
+
+original_scope_source_index :=
+  'E'    // Tag: 0x4 unsigned
+  uINDEX
+
+original_scope_end :=
+  'C'     // Tag: 0x2 unsigned
+  uLINE
+  uCOLUMN
+```
+
+The encoding scheme supports two different ways to connect a top-level `original_scope_tree` with a specific source file:
+
+  1) Use a `original_scope_source_index` item. It's `uINDEX` is an index into the `sources` array.
+  2) If `scopes` contains exactly N top-level `original_scope_trees` where N is the length of the `sources` array, then the `i`th
+     `original_scope_tree` describes `sources[i]`.
+
+The `uFLAGS` field in `original_scope_start` is a bit field defined as follows:
   * 0x1 has name
   * 0x2 has kind
   * 0x4 is stack frame
-* name: (only exists if `has name` flag is set)
-  * DATA offset into `names` field
-  * Note: This name should be shown as function name in the stack trace for function scopes.
-* kind: (only exists if `has kind` flag is set)
-  * DATA offset into `names` field
-  * Note: This offset is relative to the offset of the last `kind` or absolute if this is the first `kind`.
-  * Note: This is type of the scope.
-  * Note: JavaScript implementations should use `'global'`, `'class'`, `'function'`, and `'block'`.
-* variables:
-  * for each variable:
-    * DATA offset into `names` field for the original symbol name defined in this scope
 
-#### End Original Scope
+`original_scope_variables` is a list of indices into the `names` array of the source map JSON. The list describes the original names
+of the scope's variables.
 
-* DATA line in the original code
-  * Note: `line` is relative to the `line` of the preceding "start/end original scope" item.
-* DATA column in the original code
-  * Note: Column is always absolute.
+To reduce the number of bytes required to encode the "scopes" information, we use relative values where possible:
 
-#### Start Generated Range
+  * `sNAME` in `original_scope_start` is relative to the previous occurrence of `sNAME` (or absolute for the first).
+  * `sKIND` in `original_scope_start` is relative to the previous occurrence of `sKIND` (or absolute for the first).
+  * `uLINE` in `original_scope_start` and `original_scope_end` are relative to the previous occurrence (or absolute for the first).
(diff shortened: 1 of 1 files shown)


proposals/scopes.md
```diff
@@ -196,106 +196,179 @@ interface OriginalPosition {
...
 
-#### Start Original Scope
+```
+scopes := original_scope_list generated_range_list
+
+original_scope_list :=
+    original_scope_tree
+  | ε
+  | original_scope_list ',' original_scope_tree
+  | original_scope_list ',' ε
```
takikawa, line 221: I might be missing something, but should this grammar be something like the following?

```
original_scope_list :=
    original-scope_tree ','
  | ε
  | original_scope_list original_scope_tree ','
```

Because IIUC the `original_scope_list` should always end with a comma since `generated_range_list` comes after and is mandatory and non-empty.

szuend: That is indeed correct if we mandate generated ranges. I was actually redoing this the other day in the spec draft and came up with slightly different grammar here.

I'm not convinced yet whether we need to mandate generated ranges in the first place, or if we should allow encodings that have either scopes or ranges, or both.

If we map out all the possible combinations, the grammar would be something like this:

```
scopes :=
    original_scope_tree_list
  | top_level_item_list
  | original_scope_tree_list `,` top_level_item_list

original_scope_tree_list :=
    original_scope_tree_item
  | original_scope_tree_list `,` original_scope_tree_item

original_scope_tree_item :=
    original_scope_tree
  | ε

top_level_item_list :=
    top_level_item
  | top_level_item_list `,` top_level_item

top_level_item :=
    generated_range_tree
  | unknown_item
```

This would allow original scopes or generated ranges to be omitted. But would mandate that original scopes must come first, and after any ranges/unknown items (but not empty entries).

szuend: Ok to update the proposal text with my proposed grammar to also include unknown items?

The only top-level items I think we should mandate are original scopes. It can still make sense to translate stack traces with just original scope trees as you can use mappings to figure out the corresponding original scope even without generated ranges.

So change the `scopes` production to:

```
scopes :=
    original_scope_tree_list
  | original_scope_tree_list `,` top_level_item_list
```

proposals/scopes.md
```diff
@@ -196,106 +196,179 @@ interface OriginalPosition {
...
+Multiple top-level `generated_range_tree`s are allowed, this is especially useful when multiple bundles are straight-up concatenated.
+
+#### Original Scope Trees
+
+```
+original_scope_tree :=
+  original_scope_start
+  original_scope_variables?
+  original_scope_source_index?
+  original_scope_tree?
```
hbenl, line 238: There can be multiple `original_scope_tree` children.

proposals/scopes.md
```diff
@@ -196,106 +196,179 @@ interface OriginalPosition {
...
+
+#### Generated Range Trees
+
+```
+generated_range_tree :=
+  generated_range_start
+  generated_range_callsite?
+  generated_range_bindings?
+  generated_range_subrange_binding?
+  generated_range_tree?
```
hbenl, line 291: There can be multiple `generated_range_tree` children.

hbenl (commented):
I'd like to propose some changes:
- drop the `,` between `original_scope_tree`s and `generated_range_tree`s, I don't think they're necessary
- drop `original_scope_source_index` and mandate that there should be exactly as many original scope trees as sources

proposals/scopes.md
```diff
@@ -196,106 +196,179 @@ interface OriginalPosition {
...
+  'B'     // Tag: 0x1 unsigned
+  uFLAGS
+  uLINE
+  uCOLUMN
+  sNAME?  // Present if FLAGS<0> is set.
+  sKIND?  // Present if FLAGS<1> is set.
+
+original_scope_variables :=
+  'D'        // Tag: 0x3 unsigned
+  sVARIABLE+
```
hbenl, line 256: I think we need to add the number of variables here (but not to the bindings in generated ranges as they should always have the same number of bindings as the number of variables in the corresponding original scope).

proposals/scopes.md
```diff
@@ -196,106 +196,179 @@ interface OriginalPosition {
...
+
+`generated_range_bindings` are only valid for generated ranges that have a `sDEFINITION`. The bindings list must be equal in length as the variable list of the original scope the `sDEFINITION` references. `sBINDING+` is a list of indices into the `"names"` field of the source map JSON. Each binding is a JavaScript expression that, when evaluated, produces the **value** of the corresponding variable.
+
+`sBINDING+` indices are encoded absolute. To signify that a variable is unavailable, use the index `-1`.
+
+```
+generated_range_subrange_binding :=
+  'I'             // Tag: 0x8 unsigned
+  uVARIABLE_INDEX
+  binding_from+
```
hbenl, line 357: I think we need to add the number of subrange bindings here.

proposals/scopes.md
szuend: Nice catch, I'll update it to be a `original_scope_tree*`.

szuend:
> I'd like to propose some changes:
> 
> * drop the `,` between `original_scope_tree`s and `generated_range_tree`s, I don't think they're necessary
> * drop `original_scope_source_index` and mandate that there should be exactly as many original scope trees (or `empty_item`s) as sources (as discussed in our last meeting)

I think i dropped all references to `original_scope_source_index`. I'll double check if I missed any.
For the `,`, there might be a mistake in the grammar, but in the end we should separate every item with `,`. That includes a `,` between the last `original_scope_end` and the first `generated_range_start`.

This way, a simple decoder can just split via `,` and get individual items that way.

proposals/scopes.md
szuend: Whats the motivation? I skipped the number of variables on purpose and accept that we can't extend the `original_scope_variables` item in the future. Decoder can read VLQs until they hit the next `,`. And if we ever switch to a binary format, we'd prepend the length of the whole item either before or after the tag VLQ.

hbenl: > Decoder can read VLQs until they hit the next `,`

But in the grammar `original_scope_variables` is followed by `original_scope_source_index` or `original_scope_tree` or `original_scope_end`, without a `,` in between. We could add the `,` there instead of adding the number of variables.

hbenl:
> This way, a simple decoder can just split via `,` and get individual items that way.

Now I remembered that this scheme is important for extensibility: without a general end-of-item marker (the `,`) a decoder would not be able to determine where a new kind of item ends.

> For the `,`, there might be a mistake in the grammar, but in the end we should separate every item with `,`. That includes a `,` between the `last original_scope_end` and the `first generated_range_start`.

I think the grammar would become clearer if we expressed it as an end-of-item marker rather than a separator, i.e. the `,` should be the last token of each item in the grammar rather than a separator in each item list. And we should explicitly state that extensions must also end their new items with a `,`.

szuend:
> (we can skip the `,` at the end of fixed-length items that are part of this spec but extension items will always need to end in `,` even if they're fixed-length)

I'm against this. We don't save that many bytes and it just makes the format hard to parse. Also for original_scope_start and generated_range_start we explicitly want the `,` as an end marker so we can extend them in the future without breaking old decoders.

Whether we call the `,` a separator or an item end marker doesn't matter, in the end, they would go into the same place. The only difference is the very last item, since it would have a trailing `,`.

proposals/scopes.md
szuend: This was a mistake on my part. I updated the grammar to align more with the spec text grammar of https://szuend.github.io/ecma426-proposal-scopes/#sec-format.

proposals/scopes.md
szuend: Same reasoning as above for variables.

proposals/scopes.md
szuend: I aligned it with the latest proposal spec text, so we use `start ',' item_list? end`. And the `item_list` can be an arbitrary number of variables or trees.

In the final version of the grammar we might want to restrict `original_scope_variables` to one occurrence at most per start/end pair, but that would make it harder to read for now.

proposals/scopes.md
szuend: Same treatment as original_scope_tree (see above).

proposals/scopes.md
szuend: I updated the proposal for now with a version of the above, let me know what you think :)

proposals/scopes.md
```diff
@@ -196,106 +196,202 @@ interface OriginalPosition {
...
+  sBINDING
+  uLINE
+  uCOLUMN
+```
+
+A variable might not be available through the full generated range, or a different expression is required for parts of the generated range to retrieve a variables value. In this case a generator can use `generated_range_subrange_binding` to encode this.
+
+  * `uVARIABLE_INDEX` is an index into the corresponding original scopes' variables list. It is encoded relative inside a generated range.
+  * `binding_from` are the sub-ranges. The initial value expression for a variable is provided by the `generated_range_bindings` item. The generated position in `binding_from` is the start from which the expression `sBINDING` from `binding_from` needs to be used to retrieve the variables value instead.
+  * `sBINDING` is an index into the `"expressions"` field in the source map JSON. It is relative to previous occurrences (also relative to the last `sBINDING+` in `generated_range_bindings`)
```
hbenl, line 392:
```suggestion
  * `sBINDING` is an index into the `"names"` field in the source map JSON. It is relative to previous occurrences (also relative to the last `sBINDING+` in `generated_range_bindings`)
```

proposals/scopes.md
```diff
@@ -196,106 +196,202 @@ interface OriginalPosition {
...
+generated_range_start :=
+  'F'          // Tag: 0x5 unsigned
+  uFLAGS
+  uLINE?       // Present if FLAGS<0> is set.
+  uCOLUMN
+  sDEFINITION? // Present if FLAGS<1> is set.
+
+generated_range_end :=
+  'G'    // Tag: 0x6 unsigned
+  uLINE?
```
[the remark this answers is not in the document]
szuend: Actually, `FLAGS<0>` in `generated_range_start` only affects the presence of `uLINE` in the `generated_start_item`. Since we don't want to open up `generated_end_item` for future extension, we just check whether the end item has 1 or 2 VLQs. I'll add a sentence to make this more explicit.

proposals/scopes.md
takikawa: Sorry it took me a bit to get back to you on this, I think the new changes for including unknown items (that's just for future extension right?) and not requiring generated ranges makes sense.

I think there's a minor typo here in

```
  | original_scope_list ',' original_scope_tree
  | original_scope_list ',' ε
```

The `original_scope_list` should both be `original_scope_tree_list`.

IIUC this grammar also allows empty entries like ",,,, generated_range_tree" which I guess is harmless, but was that the intention?

proposals/scopes.md
```diff
@@ -366,38 +462,26 @@ A|    var _x = 1;
...
+  definition: scope start 1
 }
-;
-;
-;
+Bindings [message -> _m, y -> _y]
 End Generated Range C1 // B
-;
 Start Generated Range C0 { // C
   field flags: has definition, has callsite
```
takikawa, line 476: In this example, I think we can delete the "has callsite" bit which isn't used anymore right?

takikawa (approved):
Only had some minor comments, LGTM otherwise

proposals/scopes.md
szuend: You are right, `original_scope_list` should be `original_scope_tree_list`. Updated.

proposals/scopes.md
szuend: Indeed, removed!

szuend:
As both Holger and Asumu approved the PR I'm gonna go ahead and merge it. Thanks everyone!

PR #172, opened 2025-03-03
Editorial: convert to ecmarkup

nicolo-ribaudo:
This PR migrates our spec from Bikeshed to Ecmarkup, trying to keep the changes between the two versions as minimal as possible (i.e. only where we use some bikeshedisms that do not work with Ecmarkup).

I still need to make GitHub pages work, but for now you can:
1. Clone this branch
2. Run `npm ci`
3. Run `node --run build` or `node --run watch`
4. Serve the `out` folder and open it in the browser

This PR unlocks possible improvement, such as @szuend's grammar-based VLQ parsing definition, but they will be done in follow-ups.

**Preview:** https://nicolo-ribaudo.github.io/source-map/branch/ecmarkup/


spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+
+<emu-intro id="sec-intro">
+  <h1>Introduction</h1>
+  <p>This Ecma Standard defines the Source map format, used for mapping transpiled source code back to the original sources.</p>
+  <p>The source map format has the following goals:</p>
+  <ul>
+    <li>Support source-level debugging allowing bidirectional mapping</li>
+    <li>Support server-side stack trace deobfuscation</li>
+  </ul>
+  <p>The original source map format (v1) was created by Joseph Schorr for use by Closure Inspector to enable source-level debugging of optimized JavaScript code (although the format itself is language agnostic). However, as the size of the projects using source maps expanded, the verbosity of the format started to become a problem. The v2 format (Source Map Revision 2 Proposal) was created by trading some simplicity and flexibility to reduce the overall size of the source map. Even with the changes made with the v2 version of the format, the source map file size was limiting its usefulness. The v3 format is based on suggestions made by Pavel Podivilov (Google).</p>
```
takikawa, line 107: "original source" here might need `<emu-not-ref>` to avoid spurious linking

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+        ParseSourceMap (
+          _string_: a String,
+          _baseURL_: an URL,
+        ): a Decoded Source Map Record
+      </h1>
+      <dl class="header"></dl>
+      <emu-alg>
+        1. Let _json_ be Call(%JSON.parse%, *null*, « _string_ »).
+        1. If _json_ is not an Object or IsArray(_json_) is *true*, throw an error.
+        1. If HasOwnProperty(_json_, *"sections"*) is *false*, then
```
takikawa, line 377: I think this should be `is *true*, then` to take the index map branch.

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+            1. Let _offsetMappings_ be a new empty List.
+            1. For each Decoded Mapping Record _mapping_ of _decodedSection_.[[Mappings]], do
+              1. If _mapping_.[[GeneratedLine]] = 0, then
+                1. Set _mapping_.[[GeneratedColumn]] to _mapping_.[[GeneratedColumn]] + _offsetColumn_.
+              1. Set _mapping_.[[GeneratedLine]] to _mapping_.[[GeneratedLine]] + _offsetLine_.
+              1. Append _mapping_ to _offsetMappings_.
+            1. Set _sourceMap_.[[Mappings]] to the list-concatenation of _sourceMap_.[[Mappings]] and _offsetMappings_.
+            1. Set _previousOffset_ to _offset_.
+            1. Let _sortedMappings_ be a copy of _offsetMappings_, sorted in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if GeneratedPositionLessThen(_a_, _b_) is *true*.
+            1. If _sortedMappings_ is not empty, set _previousLastMappings_ to the last element of _sortedMappings_.
```
takikawa, line 907: `_previousLastMappings_` -> `_previousLastMapping_`

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+      1. If Get(_json_, *"version"*) is not *3<sub>𝔽</sub>*, optionally report an error.
+      1. Let _fileField_ be GetOptionalString(_json_, *"file"*).
+      1. Let _sourceMap_ be the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]:  « », [[Mappings]]: « » }.
+      1. Let _previousOffset_ be *null*.
+      1. Let _previousLastMapping_ be *null*.
+      1. For each _section_ of CreateListFromArrayLike(_sectionsField_), do
+        1. If _section_ is not an Object or IsArray(_section_) is *true*, optionally report an error.
+        1. Else,
+          1. Let _offset_ be Get(_section_, *"offset"*).
+          1. If _offset_ is not an Object or IsArray(_offset_) ir *true*, throw an error.
```
takikawa, line 872: `ir *true*` -> `is *true*`

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+            1. Let _sortedMappings_ be a copy of _offsetMappings_, sorted in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if GeneratedPositionLessThen(_a_, _b_) is *true*.
+            1. If _sortedMappings_ is not empty, set _previousLastMappings_ to the last element of _sortedMappings_.
+        1. Return _sourceMap_.
+    </emu-alg>
+
+    <emu-note>
+      Implementations may choose to represent index source map sections without appending the mappings together, for example, by storing each section separately and conducting a binary search.
+    </emu-note>
+
+    <emu-clause id="sec-GeneratedPositionLessThen" type="abstract operation">
```
takikawa, line 915: `GeneratedPositionLessThen` -> `GeneratedPositionLessThan` (+ uses above/below)

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+            MatchSourceMapURL (
+              _comment_: a String,
+            ): either ~NONE~ or a String
+          </h1>
+          <dl class="header"></dl>
+          <emu-alg>
+            1. Let _pattern_ be RegExpCreate(*"^[@#]\\s\*sourceMappingURL=(\\S\*?)\\s\*$"*, *""*).
+            1. Let _match_ be RegExpExec(_pattern_, _comment_).
+            1. If _match_ is not *null*, return Get(_match_, *"1"*).
+            1. Return *null*.
```
takikawa, line 1104: Should this be `Return ~NONE~`?

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+  </emu-clause>
+</emu-clause>
+
+<emu-annex id="sec-notes">
+  <h1>Notes</h1>
+
+  <emu-annex id="sec-language-neutral-mapping">
+    <h1>Language neutral stack mapping</h1>
+
+    <p>Stack tracing mapping without knowledge of the source language is not covered by this document.</p>
```
takikawa, line 1221: `<emu-not-ref>` on "covered"

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+    <li>The <dfn id="json-names"><code>names</code> field</dfn> is an optional list of symbol names which may be used by the mappings field.</li>
+    <li>The <dfn id="json-mappings"><code>mappings</code> field</dfn> is a string with the encoded mapping data (see section <emu-xref href="#sec-mappings"></emu-xref>).</li>
+    <li>The <dfn id="json-ignoreList"><code>ignoreList</code> field</dfn> is an optional list of indices of files that should be considered third party code, such as framework code or bundler-generated code. This allows developer tools to avoid code that developers likely don't want to see or step through, without requiring developers to configure this beforehand. It refers to the sources field and lists the indices of all the known third-party sources in the source map. Some browsers may also use the deprecated `x_google_ignoreList` field if `ignoreList` is not present.</li>
```
takikawa, line 304: `<emu-not-ref>` on "bundler-generated code" maybe

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+        </h1>
+        <dl class="header"></dl>
+        <emu-alg>
+          1. Let _segmentLen_ be the length of _segment_.
+          1. If _position_.[[Value]] = _segmentLen_, return *null*.
+          1. Let _first_ be ConsumeBase64ValueAt(_segment_, _position_).
+          1. Assert: _first_ &lt; 64.
+          1. If _first_ modulo 2 is 0, let _sign_ be 1.
+          1. Else, let _sign_ be -1.
+          1. Let _value_ be min(floor(_value_ / 2), 2<sup>4</sup> - 1).
```
takikawa, line 659: Should this be `floor(_first_ / 2)`?

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+          1. Let _nextShift_ be 16.
+          1. Let _currentByte_ be _first_.
+          1. Repeat, while floor(_currentByte_ / 2<sup>5</sup>) modulo 2 = 1,
+            1. If _position_.[[Value]] = _segmentLen_, throw an error.
+            1. Set _currentByte_ to ConsumeBase64ValueAt(_segment_, _position_).
+            1. Let _chunk_ be min(_value_, 2<sup>5</sup> - 1).
```
takikawa, line 665: IIUC `_value_` here should be `_currentByte_`

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+            1. Set _value_ to _value_ + _chunk_ × _nextShift_.
+            1. If _value_ &ge; 2<sup>32</sup>, throw an error.
```
takikawa, line 667: I think in the bikeshed spec this is 2^31, is 2^32 correct here?

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+            ConsumeBase64ValueAt (
+              _string_: a String,
+              _position_: a Record with a non-negative integer [[Value]] field,
+            ): a non-negative integer
+          </h1>
+          <dl class="header"></dl>
+          <emu-alg>
+            1. Assert: _position_.[[Value]] is a non-negative integer smaller than the length of _string_.
+            1. Let _char_ be the substring of _string_ from _position_ to _position_+ 1.
+            1. Assert: _char_ is a valid <emu-xref href="#sec-references-informative">base64</emu-xref> character as defined by IETF RFC 4648.
```
takikawa, line 684: Should there be a step after this one that does `Set _position_.[[Value]] to _position_.[[Value]] + 1`?

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+    <emu-clause id="sec-ParseSourceMap" type="abstract operation">
+      <h1>
+        ParseSourceMap (
+          _string_: a String,
+          _baseURL_: an URL,
+        ): a Decoded Source Map Record
+      </h1>
+      <dl class="header"></dl>
+      <emu-alg>
+        1. Let _json_ be Call(%JSON.parse%, *null*, « _string_ »).
```
szuend, line 375: Technically this would require implementors to follow exact JSON.parse semantics. And below to follow ECMAScript specific semantics (`HasOwnProperty`).

Should we instead use a "host-defined" or "implementation-defined" operation that returns a JSON `object`?

IMO the "base data types" the source map spec should operate on are the ones defined by JSON: `object`, `array`, `string`, `number`, `true`, `false` and `null`. And we assemble spec records from these.

spec.emu
```diff
@@ -0,0 +1,1319 @@
...
+          1. NOTE: The regular expression above matches the |LineTerminatorSequence| production.
+          1. Let _lastURL_ be *null*.
+          1. For each _lineStr_ in _lines_, do:
+            1. Let _line_ be StringToCodePoints(_lineStr_).
+            1. Let _position_ be 0.
+            1. Let _lineLength_ be the length of _line_.
+            1. While _position_ &lt; _lineLength_,
+              1. Let _first_ be _line_[_position_].
+              1. Set _position_ to _position_ + 1.
+              1. If _first_ is U+002F (SOLIDUS) and _position_ &lt; _lineLength_, then
```
takikawa, line 1028: I noticed sometimes U+002F is SOLIDUS like here and sometimes /. Should we make it always consistently SOLIDUS?

takikawa (approved):
This looks great, thanks for doing the conversion! I left some comments about small issues, LGTM other than those.

szuend:
This does indeed look great! I only have small concern that we tie decoding very closely to Ecma262 language semantics.

E.g. we could spec decoding of mappings as syntax directed operations over a grammar like

```html
<emu-grammar type="definition">
  Mappings :
    `;` Mappings
    MappingList Mappings?

  MappingList :
    Mapping
    MappingList `,` Mapping

  Mapping :
    GeneratedColumn
    GeneratedColumn OriginalSource OriginalLine OriginalColumn
    GeneratedColumn OriginalSource OriginalLine OriginalColumn Name

  GeneratedColumn : Vlq
  OriginalSource : Vlq
  OriginalLine : Vlq
  OriginalColumn : Vlq
  Name : Vlq
</emu-grammar>
```

then we wouldn't need to use `String.prototype.split`. 

We can do that as a follow-up though to get us started.

spec.emu
nicolo-ribaudo: Note that this was already the case, as we were using [parse JSON bytes to an Infra value](https://infra.spec.whatwg.org/#parse-a-json-string-to-an-infra-value) which internally does `Call(%JSON.parse%, *null*, « _string_ »)`.

What about adding a sentence in the new "Notational Conventions" section saying that, when our spec references ECMA-262 operations, those are to be interpreted as running in a fresh not-user-modified environment? Then whether you actually use JavaScript object or another representation is not observable (even though I think it would not be observable even if the user messes up with `Object.prototype`, because nether `JSON.parse` nor `HasOwnProperty` on the object it returns are affected by it)

szuend: The infra spec uses `%JSON.parse%` but then also immediately converts the resulting JS object back into a "infra value" (unordered map) so "callers" don't have to use ECMA-262 operations.

I'm just wondering if the spec will be more readable if we define a small `Get` and `Has` on JSON values ourselves, rather than use `HasOwnProperty` and ECMA-262 `Get` on real JS objects.

No strong opinion though, so if you and Asumu prefer to use ECMA-262 operations, I'm also fine with that.

nicolo-ribaudo:
I pushed an update that still relies on ECMA-262's data types, so that we don't have to re-define them, but it only uses its basic version of objects (https://tc39.es/ecma262/#sec-object-type), without all the special behavior that JS supports on top of it. I also define a "JSON object" and "JSON array", which we can use across the specification rather than referring to ECMA-262's object (a JSON object is more restrictive, because it guarantees that it's deeply JSON-compatible).

spec.emu
```diff
@@ -0,0 +1,1437 @@
...
+      ): a JSON value or ~missing~
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns the value associated with the specified _key_ in _object_.</dd>
+    </dl>
+    <emu-alg>
+      1. If _object_ does not have an own property with key _key_, return ~missing~.
+      1. Let _prop_ be _object_'s own property whose key is _key_.
+      1. Return _prop_.[[Value]].
```
takikawa, line 332: This might be overly nitpicky, but I think `[[Value]]` here is technically an attribute rather than field so should we do something like `Return _prop_'s [[Value]] attribute`?

takikawa:
I looked through the latest changes wrt ECMA-262 datatypes and left some comments. Otherwise the new changes LGTM.

szuend (approved):
Very nice! Thanks!

PR #177, opened 2025-03-10
Format spec.emu via 'npm run format' (emu-format)

szuend:
From a quick check the spec renders the same but I might miss something. Are there incompatibilities with emu-format and the source map spec? 

Should we add a git hook for formatting?


nicolo-ribaudo (commented):
Could you also add a CI check that the document is formatted?

nicolo-ribaudo:
These workflow run on `pull_request_target` rather than on `pull_request`. `pull_request_target` means that the workflow has push access, but it has the disadvantage that it's running the CI job as specified in the target branch (i.e. `main`) rather than in the PR branch. It will work once merged.

nicolo-ribaudo (approved):

PR #180, opened 2025-03-10
Editorial: Use grammar for VLQ and Mappings and decode via SDO

szuend:
Draft that changes VLQ and mappings to use grammar plus "syntax-directed operations" for decoding rather than a pure algorithmic approach.

Only uploaded so we have a preview for discussion.

Preview: https://szuend.github.io/source-map/branch/mappings-grammar/#sec-mappings

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -765,13 +765,13 @@
 
       <emu-grammar type="definition">
         MappingsField :
-          SegmentList
+          LineList
 
-        SegmentList :
-          Segment
-          Segment `;` SegmentList
+        LineList :
+          Line
+          Line `;` LineList
 
-        Segment :
+        Line :
           MappingList?
 
         MappingList :
@@ -849,26 +849,22 @@
         </h1>
         <dl class="header"></dl>
         <emu-grammar>
-          SegmentList :
-            Segment `;` SegmentList
-
+          LineList :
+            Line `;` LineList
+        </emu-grammar>
+        <emu-alg>
+          1. Perform DecodeMappingsField of |Line| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
+          1. Set _state_.[[GeneratedColumn]] to 0.
+          1. Perform DecodeMappingsField of |LineList| with arguments _state_, _mappings_, _names_ and _sources_.
+        </emu-alg>
+        <emu-grammar>
           MappingList :
             Mapping `,` MappingList
         </emu-grammar>
         <emu-alg>
-          1. For each child node _child_ of this Parse Node, do
-            1. If _child_ is an instance of a nonterminal, then
-              1. Perform DecodeMappingsField of _child_ with arguments _state_, _mappings_, _names_ and _sources_.
-        </emu-alg>
-        <emu-grammar>
-          Segment :
-            MappingList?
-        </emu-grammar>
-        <emu-alg>
-          1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
-          1. Set _state_.[[GeneratedColumn]] to 0.
-          1. If |MappingList| is present, then
-            1. Perform DecodeMappingsField on |MappingList| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |Mapping| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |MappingList| with arguments _state_, _mappings_, _names_ and _sources_.
         </emu-alg>
         <emu-grammar>
           Mapping :
--- a/spec.emu
+++ b/spec.emu
@@ -770,20 +770,19 @@
       <p>The mappings String must adhere to the following grammar:</p>
 
       <emu-grammar type="definition">
-        Mappings :
-          SemicolonList? Segment SemicolonList?
+        MappingsField :
+          SegmentList
+
+        SegmentList :
+          Segment
+          Segment `;` SegmentList
 
         Segment :
-          MappingList
-          MappingList SemicolonList Segment
+          MappingList?
 
         MappingList :
           Mapping
-          MappingList `,` Mapping
-
-        SemicolonList :
-          `;`
-          SemicolonList `;`
+          Mapping `,` MappingList
 
         Mapping :
           GeneratedColumn
@@ -845,9 +844,9 @@
         </table>
       </emu-table>
 
-      <emu-clause id="sec-DecodeMappingsSdo" type="sdo">
+      <emu-clause id="sec-DecodeMappingsField" type="sdo">
         <h1>
-          DecodeMappingsSdo (
+          DecodeMappingsField (
             _state_: a Decode Mapping State Record,
             _mappings_: a List of Decoded Mapping Records,
             _names_: a List of Strings,
@@ -856,28 +855,33 @@
         </h1>
         <dl class="header"></dl>
         <emu-grammar>
-          Mappings :
-            SemicolonList? Segment SemicolonList?
-
-          Segment :
-            MappingList
-            MappingList SemicolonList Segment
+          SegmentList :
+            Segment `;` SegmentList
 
           MappingList :
-            Mapping
-            MappingList `,` Mapping
+            Mapping `,` MappingList
         </emu-grammar>
         <emu-alg>
           1. For each child node _child_ of this Parse Node, do
             1. If _child_ is an instance of a nonterminal, then
-              1. Perform DecodeMappingsSdo of _child_ with arguments _state_, _mappings_, _names_ and _sources_.
+              1. Perform DecodeMappingsField of _child_ with arguments _state_, _mappings_, _names_ and _sources_.
+        </emu-alg>
+        <emu-grammar>
+          Segment :
+            MappingList?
+        </emu-grammar>
+        <emu-alg>
+          1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
+          1. Set _state_.[[GeneratedColumn]] to 0.
+          1. If |MappingList| is present, then
+            1. Perform DecodeMappingsField on |MappingList| with arguments _state_, _mappings_, _names_ and _sources_.
         </emu-alg>
         <emu-grammar>
           Mapping :
             GeneratedColumn
         </emu-grammar>
         <emu-alg>
-          1. Perform DecodeMappingsSdo of |GeneratedColumn| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |GeneratedColumn| with arguments _state_, _mappings_, _names_ and _sources_.
           1. Let _decodedMapping_ be a new DecodedMappingRecord { [[GeneratedLine]]: _state_.[[GeneratedLine]], [[GeneratedColumn]]: _state_.[[GeneratedColumn]], [[OriginalSource]]: *null*, [[OriginalLine]]: *null*, [[OriginalColumn]]: *null*, [[Name]]: *null* }.
           1. Append _decodedMapping_ to _mappings_.
         </emu-alg>
@@ -886,28 +890,17 @@
             GeneratedColumn OriginalSource OriginalLine OriginalColumn Name?
         </emu-grammar>
         <emu-alg>
-          1. Perform DecodeMappingsSdo of |GeneratedColumn| with arguments _state_, _mappings_, _names_ and _sources_.
-          1. Perform DecodeMappingsSdo of |OriginalSource| with arguments _state_, _mappings_, _names_ and _sources_.
-          1. Perform DecodeMappingsSdo of |OriginalLine| with arguments _state_, _mappings_, _names_ and _sources_.
-          1. Perform DecodeMappingsSdo of |OriginalColumn| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |GeneratedColumn| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalSource| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalLine| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalColumn| with arguments _state_, _mappings_, _names_ and _sources_.
           1. Let _source_ be _sources_[_state_.[[SourceIndex]]].
           1. If |Name| is present, then
-            1. Perform DecodeMappingsSdo of |Name| with arguments _state_, _mappings_, _names_ and _sources_.
+            1. Perform DecodeMappingsField of |Name| with arguments _state_, _mappings_, _names_ and _sources_.
             1. Let _name_ be _names_[_state_.[[NameIndex]]].
           1. Else, let _name_ be *null*.
           1. Let _decodedMapping_ be a new DecodedMappingRecord { [[GeneratedLine]]: _state_.[[GeneratedLine]], [[GeneratedColumn]]: _state_.[[GeneratedColumn]], [[OriginalSource]]: _source_, [[OriginalLine]]: _state_.[[OriginalLine]], [[OriginalColumn]]: _state_.[[OriginalColumn]], [[Name]]: _name_ }.
           1. Append _decodedMapping_ to _mappings_.
-        </emu-alg>
-        <emu-grammar>
-          SemicolonList :
-            `;`
-            SemicolonList `;`
-        </emu-grammar>
-        <emu-alg>
-          1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
-          1. Set _state_.[[GeneratedColumn]] to 0.
-          1. If |SemicolonList| is present, then
-            1. Perform DecodeMappingsSdo on |SemicolonList| with arguments _state_, _mappings_, _names_ and _sources_.
         </emu-alg>
         <emu-grammar>
           GeneratedColumn :
@@ -968,12 +961,12 @@
       <dl class="header"></dl>
       <emu-alg>
         1. Let _mappings_ be a new empty List.
-        1. Let _mappingsNode_ be the root Parse Node when parsing _rawMappings_ using |Mappings| as the goal symbol.
+        1. Let _mappingsNode_ be the root Parse Node when parsing _rawMappings_ using |MappingsField| as the goal symbol.
         1. If parsing failed, then
           1. Optionally report an error.
           1. Return _mappings_.
         1. Let _state_ be a new Decode Mapping State Record with all fields set to 0.
-        1. Perform DecodeMappingsSdo of _mappingsNode_ with arguments _state_, _mappings_, _names_ and _sources_.
+        1. Perform DecodeMappingsField of _mappingsNode_ with arguments _state_, _mappings_, _names_ and _sources_.
         1. Return _mappings_.
       </emu-alg>
     </emu-clause>
--- a/spec.emu
+++ b/spec.emu
@@ -890,10 +890,10 @@
           1. Perform DecodeMappingsSdo of |OriginalSource| with arguments _state_, _mappings_, _names_ and _sources_.
           1. Perform DecodeMappingsSdo of |OriginalLine| with arguments _state_, _mappings_, _names_ and _sources_.
           1. Perform DecodeMappingsSdo of |OriginalColumn| with arguments _state_, _mappings_, _names_ and _sources_.
-          1. Let _source_ be _sources_.[_state_.[[SourceIndex]]].
+          1. Let _source_ be _sources_[_state_.[[SourceIndex]]].
           1. If |Name| is present, then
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -275,12 +276,99 @@
...
+
+      VlqDigitList ::
+        DigitWithoutContinuationBit
+        DigitWithContinuationBit VlqDigitList
+
+      DigitWithoutContinuationBit :: one of
+        `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M` `N` `O` `P` `Q` `R`
+        `S` `T` `U` `V` `W` `X` `Y` `Z` `a` `b` `c` `d` `e` `f`
+
+      DigitWithContinuationBit :: one of
```
nicolo-ribaudo, line 296: Very nitpicky, but I would find it easier to read if there was a bigger difference between the two names. Wdyt about `TerminalDigit` and `ContinuationDigit`?

nicolo-ribaudo:
I attempted to re-express the DecodeBase64VLQ operation without using an accumulator, since the way fields are mutated in the accumulator is not super easy to keep track of (recursion + mutable state). Do you think this version would work, or am I missing something about shifting the values?

And then you use it as "let _value_ be the VLQSignedValue of _Vlq_".

**VLQSignedValue (  )**

_Vlq_ :: _VlqDigitList_
1. Let _unsigned_ be the VLQUnsignedValue of _VlqDigitList_.
2. Let _sign_ be 1.
3. If _unsigned_ modulo 2 is 1, set _sign_ to -1.
4. Return _sign_ &times; floor(_unsigned_ / 2).

**VLQUnsignedValue ( )**

_VlqDigitList_ :: _DigitWithoutContinuationBit_
1. Return DecodeBase64Digit(_DigitWithoutContinuationBit_).

_VlqDigitList_ :: _DigitWithContinuationBit_ _VlqDigitList_
1. Let _left_ be DecodeBase64Digit(_DigitWithContinuationBit_).
5. Let _right_ be the VLQUnsignedValue of _VlqDigitList_.
6. Return (_right_ - 2<sup>32</sup>) &times; 2<sup>5</sup>.

szuend:
Very nice! That looks much simpler. Although I think there is a bug: We need to slice the continuation bit off _left_ and not off `_right_`:

```
3. Return _right_ × 2 ** 5 + (_left_ - 32).
```

Alternatively we could define `DecodeBase64Digit` as an SDO that handles the continuation bit.

nicolo-ribaudo:
You are right, the `- 32` is in the wrong place :+1:

spec.emu
szuend: I like it! Done.

szuend:
With https://github.com/tc39/ecmarkup/pull/637 still in-flight, I applied Nicolos idea but via an additional nonterminal indirection.

spec.emu
```diff
@@ -275,12 +276,97 @@
...
+      OneOfTerminalDigit :: one of
+        `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M` `N` `O` `P` `Q` `R`
+        `S` `T` `U` `V` `W` `X` `Y` `Z` `a` `b` `c` `d` `e` `f`
+
+      OneOfContinuationDigit :: one of
+        `g` `h` `i` `j` `k` `l` `m` `n` `o` `p` `q` `r` `s` `t` `u` `v` `w` `x`
+        `y` `z` `0` `1` `2` `3` `4` `5` `6` `7` `8` `9` `+` `/`
+    </emu-grammar>
+
+    <emu-note>
```
nicolo-ribaudo, line 307:
```suggestion
    <emu-note type="editor">
```

Ecmarkup supports "editorial notes", to write notes on how the spec is written rather than on its meaning.

spec.emu
```diff
@@ -275,12 +276,97 @@
...
+      <h1>VLQSignedValue ( ): an integer</h1>
+      <dl class="header"></dl>
+      <emu-grammar>
+        Vlq :: VlqDigitList
+      </emu-grammar>
+      <emu-alg>
+        1. Let _unsigned_ be the VLQUnsignedValue of |VlqDigitList|.
+        1. If _unsigned_ modulo 2 = 1, let _sign_ be -1.
+        1. Else, let _sign_ be 1.
+        1. Return _sign_ × floor(_unsigned_ / 2).
```
nicolo-ribaudo, line 321: I just realized that we need to preserve step 11 of https://tc39.es/ecma426/#sec-DecodeBase64VLQ: _If value is 0 and sign is -1, return -2<sup>31</sup>._

spec.emu
```diff
@@ -275,12 +276,97 @@
...
+    <emu-clause id="sec-VLQUnsignedValue" type="sdo">
+      <h1>VLQUnsignedValue ( ): an non-negative integer</h1>
+      <dl class="header"></dl>
+      <emu-grammar>
+        VlqDigitList ::
+          TerminalDigit
+      </emu-grammar>
+      <emu-alg>
+        1. Return VLQUnsignedValue of |TerminalDigit|.
+      </emu-alg>
```
nicolo-ribaudo, line 334: This is actually not needed, as the default behavior for SDOs is to apply itself on the only child: https://tc39.es/ecma262/#sec-algorithm-conventions-syntax-directed-operations (from "Unless explicitly specified otherwise...").

spec.emu
szuend: Done!

spec.emu
szuend: Done.

spec.emu
szuend: I missed the edit and added the check as it was previously. I also added back in the checks for >= 2 ** 31, both for the signed and unsigned case. PTAL if this works as I intend it to xD

nicolo-ribaudo (commented):
Second review pass. The VLQs look good now, and I finally reviewed the mappings definitions.

I would prefer this slightly different approach, based on the "informal" understanding that mappings contains one or more segments, one per line, separated by semicolons. A segment contains zero or more mappings, separated by commas. This avoids having to think about the "what if there are consecutive semicolons?" case, since they are just consecutive segments.

```
Mappings :
  Segment
  Segment `;` Mappings

Segment :
  MappingList?

MappingList :
  Mapping
  Mapping `,` MappingList
```

Note that the above definitions makes `"mappings": ""` valid, which matches the current spec.

_DecodeMappingsSdo_ would then become
**DecodeMappingsSdo ( ... )**

_Mappings : Segment `;` Mappings_
_MappingList : Mapping `,` MappingList_
1. For each child node _child_ of this Parse Node, do ... (same as now)

_Segment : MappingList?_
1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
2. Set _state_.[[GeneratedColumn]] to 0.
3. If _MappingList_ is present, perform DecodeMappingsSdo of _MappingList_ with arguments ...

_Mapping : GeneratedColumn_ (same as now)
_Mapping :  GeneratedColumn OriginalSource OriginalLine OriginalColumn Name?_ (same as now)

szuend:
Thanks for the new grammar, I love it!

> Maybe can we also rename _Mappings_ to _SegmentList_, or _MappingsSegmentList_? It's a bit weird that "mappings" is a list of segments, and "segment" is a list of mappings, even though that's how we always referred to it 😅

IMO it would be nice to have consistent naming of the goal symbols w.r.t. to how the field is named in the source map JSON. What about renaming _Mappings_ to _SegmentList_ but then add a goal symbol _MappingsField_ : _SegmentList_?

nicolo-ribaudo:
> What about renaming Mappings to SegmentList but then add a goal symbol MappingsField : SegmentList?

Sounds good 👍

szuend:
Changed the grammar as per your suggestion. Also renamed the SDO to `DecodeMappingsField`, which makes more sense. The "Sdo" suffix was only there because I couldn't think of anything better at the time.

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+            <td>[[NameIndex]]</td>
+            <td>a non-negative integer</td>
+          </tr>
+        </table>
+      </emu-table>
+
+      <emu-clause id="sec-DecodeMappingsField" type="sdo">
+        <h1>
+          DecodeMappingsField (
+            _state_: a Decode Mapping State Record,
```
takikawa, line 850: Very minor: this doesn't get hyperlinked because the variant specified above is plural but this isn't

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+          1. If _state_.[[GeneratedColumn]] &lt; 0, optionally report an error.
+        </emu-alg>
+        <emu-grammar>
+          OriginalSource :
+            Vlq
+        </emu-grammar>
+        <emu-alg>
+          1. Let _relativeSourceIndex_ the VLQSignedValue of |Vlq|.
+          1. Set _state_.[[SourceIndex]] to _state_.[[SourceIndex]] + _relativeSourceIndex_.
+          1. If _state_.[[SourceIndex]] &lt; 0 or _state.[[SourceIndex]] ≥ the number of elements of _sources_, optionally report an error.
```
takikawa, line 921: Minor typo: the second `_state_` is missing the end `_`.

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+        <emu-grammar>
+          Mapping :
+            GeneratedColumn OriginalSource OriginalLine OriginalColumn Name?
+        </emu-grammar>
+        <emu-alg>
+          1. Perform DecodeMappingsField of |GeneratedColumn| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalSource| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalLine| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |OriginalColumn| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Let _source_ be _sources_[_state_.[[SourceIndex]]].
```
takikawa, line 897: It's only in an error case, but there's a small semantic difference between the old version and this I think. If the source index is invalid, the old algorithm kept the source as `null` while this returns `undefined` in JS or out-of-bounds lookup in general. (original line/column are also kept `null` when they are negative)

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+          1. If _state_.[[OriginalColumn]] &lt; 0, optionally report an error.
+        </emu-alg>
+        <emu-grammar>
+          Name :
+            Vlq
+        </emu-grammar>
+        <emu-alg>
+          1. Let _relativeName_ be the VLQSignedValue of |Vlq|.
+          1. Set _state_.[[NameIndex]] to _state_.[[NameIndex]] + _relativeName_.
+          1. If _state_.[[NameIndex]] &lt; 0 or _state_[[NameIndex]] ≥ the number of elements of _sources_, optionally report an error.
```
jridgewell, line 948:
```suggestion
          1. If _state_.[[NameIndex]] &lt; 0 or _state_.[[NameIndex]] ≥ the number of elements of _sources_, optionally report an error.
```

jridgewell (approved):

spec.emu
szuend: Thanks! The &lt;dfn&gt; had a typo.

spec.emu
szuend: Fixed :)

spec.emu
szuend: Yeah there is some bound checks missing. Same for `names`.

I'm not sure about `undefined` though. We are using the [List specification type](https://tc39.es/ecma262/#sec-list-and-record-specification-type) here which does not actually spell out what is returned for out-of-bounds accesses, so it's even worse since its undefined behavior.

I'll add some bound checks.

spec.emu
```diff
@@ -276,6 +276,95 @@
...
+
+    <emu-clause id="sec-VLQUnsignedValue" type="sdo">
+      <h1>VLQUnsignedValue ( ): an non-negative integer</h1>
+      <dl class="header"></dl>
+      <emu-grammar>
+        Vlq :: VlqDigitList
+      </emu-grammar>
+      <emu-alg>
+        1. Let _value_ be the VLQUnsignedValue of |VlqDigitList|.
+        1. If _value_ is ≥ 2<sup>31</sup>, throw an error.
```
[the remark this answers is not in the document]
szuend: I don't think this is the case. This algorithm is defined for the grammar rule `Vlq :: VlqDigitList`, which means it is not invoked at all for `VLQSignedValue` (since that is also defined on `Vlq :: VlqDigitList`.

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+        <emu-grammar>
+          SegmentList :
+            Segment `;` SegmentList
+
+          MappingList :
+            Mapping `,` MappingList
+        </emu-grammar>
+        <emu-alg>
+          1. For each child node _child_ of this Parse Node, do
+            1. If _child_ is an instance of a nonterminal, then
```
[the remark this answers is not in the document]
szuend: I adapted this from the ECMAScript spec that has a couple instances of this (e.g. [Contains](https://tc39.es/ecma262/#sec-static-semantics-contains)).

Since we only have two productions, we could spell them also out explicitly:

```html
<emu-grammar>
  SegmentList :
    Segment `;` SegmentList
</emu-grammar>
<emu-alg>
  1. Perform DecodeMappingsField of |Segment| ...
  1. Perform DecodeMappingsFIeld of |SegmentList| ...
</eum-alg>
```

And same for the `MappingList` production.

spec.emu
```diff
@@ -676,134 +765,210 @@
...
+              1. Perform DecodeMappingsField of _child_ with arguments _state_, _mappings_, _names_ and _sources_.
+        </emu-alg>
+        <emu-grammar>
+          Segment :
+            MappingList?
+        </emu-grammar>
+        <emu-alg>
+          1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
```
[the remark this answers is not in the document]
szuend: You are right.

I think it might be sufficient to switch the statements around. Perform `DeocdeMappingsField` first, then increment the line.

Alternatively we could also move the increment to the `SegemtnList` production. I'd find that even clearer:

```
<emu-grammar>
  SegmentList :
    Segment `;` SegmentList
</emu-grammar>
<emu-alg>
  1. Perform DecodeMappingsField on |Segment| ...
  1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
  1. Set _state_.[[GeneratedColumn]] to 0.
  1. Perform DecodeMappingsField on |SegmentList| ...
</emu-alg>
```

@nicolo-ribaudo any preference?

spec.emu
szuend: I rewrote the algorithm here to do the actual validation and set the fields explicitly to *null*. PTAL.

szuend: I also filed #184. It seems some implementations (e.g. Chrome) behave as I wrote the spec initially but others (e.g. Firefox) behave as the existing spec says.

spec.emu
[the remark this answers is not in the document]
jridgewell: `VLQSignedValue` calls `VLQUnsignedValue` as its first step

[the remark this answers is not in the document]
szuend: Yes, you are indeed right.

`VLQSignedValue` calls `VLQUnsignedValue` as its first step but on the `|VlqDigitList|` parse node, not the `|Vlq|` parse node. This means signed values are fine, but the unsiged ones would be truncated.

spec.emu
```diff
@@ -276,6 +276,95 @@
       <emu-note>That means that "A" (`LATIN CAPITAL LETTER A`) measures as 1 code unit, and "🔥" (`FIRE`) measures as 2 code units. Source maps for other content types may diverge from this.</emu-note>
     </dd>
   </dl>
+
+  <emu-clause id="sec-vlq-grammar">
```
nicolo-ribaudo, line 280: Ideally "terms and definitions" should have "simple" definitions, easily expressible as "name -> meaning", like for a human-language dictionary.

I would prefer to move this emu-clause to the top level, remove the existing _base64 VLQ_ `<dfn>` from "terms and definitions", and use it as an introductory paragraph for this new section, including the examples.

spec.emu
```diff
@@ -276,6 +276,95 @@
...
+      OneOfTerminalDigit :: one of
+        `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M` `N` `O` `P` `Q` `R`
+        `S` `T` `U` `V` `W` `X` `Y` `Z` `a` `b` `c` `d` `e` `f`
+
+      OneOfContinuationDigit :: one of
+        `g` `h` `i` `j` `k` `l` `m` `n` `o` `p` `q` `r` `s` `t` `u` `v` `w` `x`
+        `y` `z` `0` `1` `2` `3` `4` `5` `6` `7` `8` `9` `+` `/`
+    </emu-grammar>
+
+    <emu-note type="editor">
```
nicolo-ribaudo, line 307: Don't forget to do this after rebasing.

spec.emu
```diff
@@ -276,6 +276,95 @@
...
+      <emu-grammar>
+        Vlq :: VlqDigitList
+      </emu-grammar>
+      <emu-alg>
+        1. Let _unsigned_ be the VLQUnsignedValue of |VlqDigitList|.
+        1. If _unsigned_ modulo 2 = 1, let _sign_ be -1.
+        1. Else, let _sign_ be 1.
+        1. Let _value_ be floor(_unsigned_ / 2).
+        1. If _value_ is 0 and _sign_ is -1, return -2<sup>31</sup>.
+        1. If _value_ is ≥ 2<sup>31</sup>, throw an error.
```
nicolo-ribaudo, line 323: - When _sign_ is -1, 2<sup>31</sup> is actually a supported value (the ranges goes from -2<sup>31</sup> to 2<sup>32</sup>-1).
- Given that we know that _unsigned_ is less than 2<sup>32</sup> (otherwise VLQUnsignedValue would have thrown), we don't actually need this check. It can be an assertion.

nicolo-ribaudo, line 323: I would maybe add a note somewhere that we need this check because we are calling VLQUnsignedValue on |VlqDigitList| rather than on |Vlq|, so we are skipping VLQUnsignedValue's validation.

Or we call VLQUnsignedValue on |Vlq|, and replace this check with an assertion.

nicolo-ribaudo (commented):
I left minor some comments, but the rest looks good to me. I'll approve once rebased.

nicolo-ribaudo:
Oh also, per the last call we had, lets add a "Grammar notation" section in "Notational conventions". It should encourage readers to check out https://tc39.es/ecma262/#sec-grammar-notation, mentioning explicitly that:
- the terminal symbols of our grammars are individual code points (i.e. they are more like the lexical and regexp grammars of ECMA-262, than like the syntactic grammar)
- we do not use https://tc39.es/ecma262/#sec-grammatical-parameters and https://tc39.es/ecma262/#sec-lookahead-restrictions, which keeps our grammar simpler.

spec.emu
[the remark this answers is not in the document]
szuend: As Justin suggested, I renamed "Segment" to "List". I slightly prefer the approach where we increment the line in the SDO of ``Line `;` LineList``, as the semicolon is directly there.

spec.emu
szuend: Done

spec.emu
szuend: Added a note. Otherwise we'd have to turn `VLQSignedValue` into an abstract op and pass in a |Vlq| parse node.

spec.emu
szuend: Done, I renamed the top level section to "base64 VLQ".

szuend:
> Oh also, per the last call we had, lets add a "Grammar notation" section in "Notational conventions".

Done. Might need some tweaking to the wording.

spec.emu
```diff
@@ -236,6 +236,15 @@
       <p>An implementation can choose different behaviors for different optional errors.</p>
     </emu-clause>
   </emu-clause>
+
+  <emu-clause id="sec-grammar-notation">
+    <h1>Grammar Notation</h1>
+    <p>This specification follows the same grammar notation convention as defined by <a href="https://tc39.es/ecma262/#sec-grammar-notation">ECMA-262 (Grammar Notation)</a>, with the following caveats:</p>
+    <ul>
+      <li>The terminal symbols of grammars in this specification are individual code points. They are more like ECMA-262 lexical and regexp grammars than the ECMA-262 syntactic grammar.</li>
+      <li>This specification does not use <a href="https://tc39.es/ecma262/#sec-grammatical-parameters">grammatical parameters</a> or <a href="https://tc39.es/ecma262/#sec-lookahead-restrictions">lookahead restrictions</a>, which keeps the grammars simpler.</li>
```
nicolo-ribaudo, line 245:
```suggestion
      <li>The terminal symbols of grammars defined in this specification are individual code points. This is similar to ECMA-262's <a href="https://tc39.es/ecma262/#sec-lexical-and-regexp-grammars>lexical grammar</a>, rather than to ECMA-262's <a href="https://tc39.es/ecma262/#sec-syntactic-grammar">syntactic grammar</a>.</li>
      <li>This specification does not use <a href="https://tc39.es/ecma262/#sec-grammatical-parameters">grammatical parameters</a> or <a href="https://tc39.es/ecma262/#sec-lookahead-restrictions">lookahead restrictions</a>, to reduce the grammar definitions complexity.</li>
```

szuend: Much appreciated!

PR #183, opened 2025-03-14
Fix sourceRoot decoding

jridgewell:
This matches impl behavior for `sourceRoot`. If set, it's treated as a path, and if it does not end with a `/`, it is appended with one:

- [Chrome](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.ts;l=486-492;drc=6ca4ad9f4aac1de91c8b5095b8fdb558ad9e7c5e)
- [Mozilla](https://github.com/mozilla/source-map/blob/3cb92cc3b73bfab27c146bae4ef2bc09dbb4e5ed/lib/util.js#L440)
- [trace-mapping](https://github.com/jridgewell/trace-mapping/blob/5a658b10d9b6dea9c614ff545ca9c4df895fee9e/src/resolve.ts#L10-L13)

I think the confusion here came from `baseURL` (the map file's URL), which **is** filename trimmed.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1080,9 +1080,8 @@
         1. Let _sourcesContentCount_ be the the number of elements in _sourcesContent_.
         1. Let _sourceUrlPrefix_ be *""*.
         1. If _sourceRoot_ ≠ *null*, then
-          1. If _sourceRoot_ contains the code point U+002F (SOLIDUS), then
-            1. Let _idx_ be the index of the last occurrence of U+002F (SOLIDUS) in _sourceRoot_.
-            1. Set _sourceUrlPrefix_ to the substring of _sourceRoot_ from 0 to _idx_ + 1.
+          1. If _sourceRoot_ ends with the code point U+002F (SOLIDUS), then
+            1. Set _sourceUrlPrefix_ to _sourceRoot_.
           1. Else,
             1. Set _sourceUrlPrefix_ to the string-concatenation of _sourceRoot_ and *"/"*.
         1. Let _index_ be 0.


spec.emu
```diff
@@ -895,10 +895,10 @@
         1. Let _sourcesContentCount_ be the the number of elements in _sourcesContent_.
         1. Let _sourceUrlPrefix_ be *""*.
         1. If _sourceRoot_ ≠ *null*, then
-          1. If _sourceRoot_ contains the code point U+002F (SOLIDUS), then
-            1. Let _idx_ be the index of the last occurrence of U+002F (SOLIDUS) in _sourceRoot_.
-            1. Set _sourceUrlPrefix_ to the substring of _sourceRoot_ from 0 to _idx_ + 1.
+          1. If _sourceRoot_ ends with the code point U+002F (SOLIDUS), then
+            1. Set _sourceUrlPrefix_ to _sourceRoot_.
           1. Else, set _sourceUrlPrefix_ to the string-concatenation of _sourceRoot_ and *"/"*.
+        1. TODO: Strip everything after last `/` from _baseURL_'s string representation.
```
nicolo-ribaudo, line 901: This is already done when we use the base URL, as it's doing the equivalent of `new URL(source, baseURL)` which removed the last segment from `baseURL` (e.g. `new URL("./foo.js", "http://example.com/folder/bar.js")` is `http://example.com/folder/foo.js`).

jridgewell: Removed.

jridgewell:
```js
const baseUrl = 'https://google.com/foo/entrypoint.js.map';
const sourceRoot = 'bar/baz';
const sourceUrl = 'source.js';
```

Under the old spec, this would produce: `https://google.com/foo/bar/source.js`, where the new spec will produce `https://google.com/foo/bar/baz/source.js`.

The filename is stripped from the `baseUrl`, but it is never striped from `sourceRoot`. If source root exists, and doesn't end with a `/`, one is appended to it.

szuend (approved):
I can confirm that the Chrome DevTools implementation [behaves](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.test.ts;l=605-608;drc=aa51ce909dc8b992cdecc4a1b449b43684461585) as this PR suggests

takikawa (approved):
I double checked the source map library for Firefox as well to see if it's implementation matched. LGTM

nicolo-ribaudo:
@jridgewell Could you rebase so we can merge this?

jridgewell:
Done.

PR #189, opened 2025-03-31
Fix missing quote in spec.emu

nicolo-ribaudo:

PR #191, opened 2025-03-31
Fix CI linting

nicolo-ribaudo:
See https://github.com/tc39/ecma426/pull/189, tested at https://github.com/nicolo-ribaudo/source-map/pull/2.

The problem was that we were linting on the `main` spec.emu, rather than the PR, because we are running on `pull_request_target`. This change gets the spec files from the PR, so that we are actually linting them.


nicolo-ribaudo:
Probably this also fixes the fact that the `tc39.es/ecma426/pr/191`-style previews show `main` rather than the PR contents.

PR #193, opened 2025-04-01
Support linting the spec

nicolo-ribaudo:
Ecmarkup supports linting, to enforce a consistent style and to prevent simple bugs (e.g. using undeclared variables). We didn't originally enable it, because of many false positives about completion records (which we do not use).

This PR patches ecmarkup to suppress false positives, enables linting, and fixes all the valid linting issues.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -268,7 +268,7 @@
 
     <dt><dfn id="sec-terms-and-definitions-colun">column</dfn></dt>
     <dd>
-      <p>zero-based indexed offset within a line of the generated code, computed as UTF-16 code units for JavaScript and CSS source maps, and as byte indeces in the binary content (represented as a single line) for WebAssembly source maps.</p>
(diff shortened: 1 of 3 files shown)


spec.emu
```diff
@@ -269,7 +268,7 @@
 
     <dt><dfn id="sec-terms-and-definitions-colun">column</dfn></dt>
     <dd>
-      <p>zero-based indexed offset within a line of the generated code, computed as UTF-16 code units for JavaScript and CSS source maps, and as byte indexes in the binary content (represented as a single line) for WebAssembly source maps.</p>
+      <p>zero-based indexed offset within a line of the generated code, computed as UTF-16 code units for JavaScript and CSS source maps, and as byte indeces in the binary content (represented as a single line) for WebAssembly source maps.</p>
```
szuend, line 271: `indices`? xD

szuend (approved):

spec.emu
nicolo-ribaudo: The rule with linters is: do not fight back, accept it and don't ask questions :P

(Also, in JS `Object.getOwnPropertyNames(RegExp.prototype)` has a property called `hasIndeces` 😉)

PR #194, opened 2025-04-07
Editorial: Add 'Position' and 'Original Position' records

szuend:
This PR adds the 'Position' record consisting of a line plus column and a 'Original Position' record, consisting of a source index, line and column.

The mapping record thus changes to contain one generated position record and one original position record. This makes the intention of a mapping a bit more clear in the spec: e.g. a [[OriginalLine]] that is null while [[OriginalColumn]] is present, does not make a lot of sense (while possible to achieve in the encoding).

The PR also provides a general comparison function and replaces the adhoc "LessThan" mapping comparison in the index decoding algorithm.

We'll also use the 'Position' record in the spec text of the "scopes" proposal.

Preview: https://tc39.es/ecma426/pr/194/

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -500,11 +500,11 @@
         </thead>
         <tr>
           <td>[[Line]]</td>
-          <td>a Number</td>
+          <td>a non-negative integral Number</td>
         </tr>
         <tr>
           <td>[[Column]]</td>
-          <td>a Number</td>
+          <td>a non-negative integral Number</td>
         </tr>
       </table>
     </emu-table>
@@ -523,15 +523,15 @@
         </thead>
         <tr>
           <td>[[SourceIndex]]</td>
-          <td>a Number</td>
+          <td>a non-negative integral Number</td>
         </tr>
         <tr>
           <td>[[Line]]</td>
-          <td>a Number</td>
+          <td>a non-negative integral Number</td>
         </tr>
         <tr>
           <td>[[Column]]</td>
-          <td>a Number</td>
+          <td>a non-negative integral Number</td>
         </tr>
       </table>
     </emu-table>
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (approved):

spec.emu
```diff
@@ -484,6 +484,81 @@
...
+      <table>
+        <thead>
+          <tr>
+            <th>Field Name</th>
+            <th>Value Type</th>
+          </tr>
+        </thead>
+        <tr>
+          <td>[[Line]]</td>
+          <td>a Number</td>
```
takikawa, line 503: Just for clarity, what do you think about putting "a non-negative integral Number" here and below?

spec.emu
```diff
@@ -895,27 +959,21 @@
...
-          1. Else, let _originalLine_ be _state_.[[OriginalLine]].
           1. Perform DecodeMappingsField of |OriginalColumn| with arguments _state_, _mappings_, _names_ and _sources_.
-          1. If _state_.[[OriginalColumn]] &lt; 0, then
+          1. If _state_.[[SourceIndex]] &lt; 0 or _state_.[[SourceIndex]] ≥ the number of elements of _sources_ or _state_.[[OriginalLine]] &lt; 0 or _state_.[[OriginalColumn]] &lt; 0, then
             1. Optionally report an error.
-            1. Let _originalColumn_ be *null*.
-          1. Else, let _originalColumn_ be _state_.[[OriginalColumn]].
+            1. Let _originalPosition_ be *null*.
+          1. Else,
+            1. Let _originalPosition_ be a new Original Position Record { [[SourceIndex]]: _sources_[_state_.[[SourceIndex]]], [[Line]]: _state_.[[OriginalLine]], [[Column]]: _state_.[[OriginalColumn]] }.
```
takikawa, line 970: Should this be `[[SourceIndex]]: _state_.[[SourceIndex]]` here instead? (since it's the index and not the record now)

takikawa (approved):
LGTM modulo some minor comments

spec.emu
szuend: Good point, done!

spec.emu
szuend: Great catch! I fixed it for now as you suggested, but do you think that the "Original Position Record" should just keep an index as it is now, or should it resolve to the actual "Decoded Source Record"?

takikawa: Hmm, that's a good question, I'm not sure. I think it could be made to work with an index too but actually it does require another change, which is that for index maps we would need to renumber the indices (since they are relative to the sub-source map).

@nicolo-ribaudo do you have any opinion on this?

nicolo-ribaudo: I don't have a strong preference in either direction.

szuend:
I rebased and made the change to "Original Position Record". Maybe have one last look if the rebase was done correctly before merging (fixed a warning about a trailing space as a drive-by). Thanks!

nicolo-ribaudo (approved):
Still looks good to me

PR #195, opened 2025-04-09
Add algorithms for retrieving original & generated positions

takikawa:
This PR is an initial attempt at specifying how we can use Decoded Source Map Records to do position lookups:

  * Adds a step to sort the mappings before they are added into the record, which helps simplify the spec algorithms
    - This allows some refactoring for index map handling
  * Adds algorithms for looking up original and generated positions
    - These currently return new types of position records but it could also make sense to just return the mapping / list of mappings

Specifying something like this helps proposals like the range mapping proposal that affect or change how mappings should be interpreted.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -608,6 +608,7 @@
         1. Let _sources_ be DecodeSourceMapSources(_baseURL_, _sourceRootField_, _sourcesField_, _sourcesContentField_, _ignoreListField_).
         1. Let _namesField_ be GetOptionalListOfStrings(_json_, *"names"*).
         1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_).
+        1. Sort _mappings_ in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if GeneratedPositionLessThan(_a_, _b_) is *true*.
         1. Let _sourceMap_ be the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_ }.
       </emu-alg>
 
@@ -698,6 +699,21 @@
                 <!-- TODO: Should this step be removed, so that the list only contains numbers? -->
               1. Append *null* to _list_.
           1. Return _list_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-GeneratedPositionLessThan" type="abstract operation">
+        <h1>
+          GeneratedPositionLessThan (
+            _a_: a Decoded Mapping Record,
+            _b_: a Decoded Mapping Record,
+          ): a Boolean
+        </h1>
+        <dl class="header"></dl>
+        <emu-alg>
+          1. If _a_.[[GeneratedLine]] &lt; _b_.[[GeneratedLine]], return *true*.
+          1. If _a_.[[GeneratedLine]] = _b_.[[GeneratedLine]] and _a_.[[GeneratedColumn]] &lt; _b_.[[GeneratedColumn]], return *true*.
+          1. Return *false*.
         </emu-alg>
       </emu-clause>
     </emu-clause>
@@ -1198,29 +1214,13 @@
               1. Append _mapping_ to _offsetMappings_.
             1. Set _sourceMap_.[[Mappings]] to the list-concatenation of _sourceMap_.[[Mappings]] and _offsetMappings_.
             1. Set _previousOffset_ to _offset_.
-            1. Let _sortedMappings_ be a copy of _offsetMappings_, sorted in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if GeneratedPositionLessThan(_a_, _b_) is *true*.
-            1. If _sortedMappings_ is not empty, set _previousLastMapping_ to the last element of _sortedMappings_.
+            1. If _offsetMappings_ is not empty, set _previousLastMapping_ to the last element of _offsetMappings_.
         1. Return _sourceMap_.
     </emu-alg>
 
     <emu-note>
       Implementations may choose to represent index source map sections without appending the mappings together, for example, by storing each section separately and conducting a binary search.
     </emu-note>
-
-    <emu-clause id="sec-GeneratedPositionLessThan" type="abstract operation">
-      <h1>
-        GeneratedPositionLessThan (
-          _a_: a Decoded Mapping Record,
-          _b_: a Decoded Mapping Record,
-        ): a Boolean
-      </h1>
-      <dl class="header"></dl>
-      <emu-alg>
-        1. If _a_.[[GeneratedLine]] &lt; _b_.[[GeneratedLine]], return *true*.
-        1. If _a_.[[GeneratedLine]] = _b_.[[GeneratedLine]] and _a_.[[GeneratedColumn]] &lt; _b_.[[GeneratedColumn]], return *true*.
-        1. Return *false*.
-      </emu-alg>
-    </emu-clause>
   </emu-clause>
 </emu-clause>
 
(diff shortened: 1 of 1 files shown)


takikawa:
I have some changes to this that I'll push after #194 is merged based on the recent TG4 discussion (adding a LUB variant of lookup for example).

takikawa:
I finally made another revision to this, with the PR rebased on the latest spec and changing how the operations work a bit.

Instead of providing multiple versions of `GetOriginalPosition`, the current PR just defines a single version that returns an exact match or the greatest lower bound. I believe the three major browser devtools all provide this, and while some provide additional versions of the operation, the spec doesn't preclude that either.

For `GetGeneratedPositions` there is less agreement between implementations. So this PR defines a version `GetGeneratedPositionsForOriginalLine` which returns all the reverse mappings that match the given original source + line. Some implementations do match on the column in some way too, but you can filter the output of this operation to get that. The idea is that this is a baseline that can be used to derive the behaviors in all the implementations.

---

For additional context, I summarized the browser devtools behavior for these two operations:

Getting the original position:
- Chrome: [`findEntry`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.ts;l=240;drc=9e10d1e251d16509bddbd360bf5969023e31c8d9): takes line, column, and optional inlineFrameIndex arguments. Finds equal or greatest lower bound item.
- Firefox: [`originalPositionFor`](https://github.com/mozilla/source-map/blob/master/lib/source-map-consumer.js#L478): takes line, column, and bias arguments. Finds equal or lub/glb based on the bias arguments.
  * `jridgewell/trace-mapping` also implements the [same API](https://github.com/jridgewell/sourcemaps/blob/main/packages/trace-mapping/src/trace-mapping.ts#L197)
  * Both lub/glb are [actually used](https://searchfox.org/mozilla-central/source/devtools/client/shared/source-map-loader/source-map.js#345), since for example you'd want lub search if you are searching at a fixed column=0 on multiple lines.
- WebKit: [`findOriginalPosition`](https://searchfox.org/wubkat/source/Source/WebInspectorUI/UserInterface/Models/SourceMap.js#312): takes line, column arguments. Returns equal or glb.

Getting generated positions:
- Chrome: [`findReverseEntries`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/sdk/SourceMap.ts;drc=9e10d1e251d16509bddbd360bf5969023e31c8d9;l=356): takes source, line, column. Finds equal or glb reverse mapping and then returns a list of mappings with that original position.
- Firefox: [`allGeneratedPositionsFor`](https://github.com/mozilla/source-map/blob/master/lib/source-map-consumer.js#L388): takes source, line, column. Finds either all exact matches (if column passed) or all mappings on source+line if column is null.
- WebKit: doesn't implement an equivalent, but there is a [single-mapping version](https://searchfox.org/wubkat/source/Source/WebInspectorUI/UserInterface/Models/SourceMap.js#338). It returns the first column position on the matching line.

takikawa:
There's a rendered preview of the PR [here](https://takikawa.github.io/source-map-spec/#sec-operations-on-source-map-records).

spec.emu
```diff
@@ -1539,6 +1539,75 @@ return lastURL;</code></pre>
...
+        1. Let _originalPosition_ be _mapping_.[[OriginalPosition]].
+        1. If _originalPosition_ ≠ *null*, then
+          1. If _originalPosition_.[[Line]] is equal to _originalLine_, then
+            1. Append _mapping_.[[GeneratedPosition]] to _generatedPositions_.
+      1. Return _generatedPositions_.
+    </emu-alg>
+
+    <emu-clause id="sec-ComputeReverseMappings" type="abstract operation">
+      <h1>
+        ComputeReverseMappings (
```
jridgewell, line 1592: Does this need to be a separate AO? We could just filter by `Decoded Mapping Record` and `originalLine` all at the same time.

spec.emu
```diff
@@ -1539,6 +1539,75 @@ return lastURL;</code></pre>
...
+          1. Return _closest_.
+        1. Else,
+          1. Set _closest_ to _mapping_.
+      1. Return _closest_.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-GetGeneratedPositionsForOriginalLine" type="abstract operation">
+    <h1>
+      GetGeneratedPositionsForOriginalLine (
```
jridgewell, line 1571: `trace-mapping` [also has](https://github.com/jridgewell/sourcemaps/blob/c780a904904048808f6043969dc1b87fc5b361d1/packages/trace-mapping/src/trace-mapping.ts#L235-L253) a `generatedPositionFor`/`allGeneratedPositionFor` and match Chrome's behavior. I could add a column optional variation, but I'm curious what use case it solves.

spec.emu
takikawa: I don't think it has to be currently, the motivation for factoring out originally was that I intended to call the helper from other operations as well (if the spec has an operation that returns a single generated mapping for example).

takikawa:
I updated this PR to just add the original position operation, since that's a more obvious starting point and lets other PRs build on it.

One thing that's unresolved is what to do when there are duplicate mappings at the same location. We could change the algorithm to return an arbitrary mapping when picking the closest element:

Current (pick last closest):
```
        1. Else,
          1. Set _closest_ to _mapping_.
```

Arbitrary:
```
        1. Else,
          1. If the result of performing ComparePositions(_mapping_.[[GeneratedPosition]], _closest_) is ~equal~, then
            1. Set _closest_ to an implementation-defined choice of _closest_ or _mapping_.
          1. Else, set _closest_ to _mapping_.
```

Any opinion on this? (or other options)

spec.emu
```diff
@@ -1539,6 +1539,33 @@ return lastURL;</code></pre>
...
+
+  <p>After decoding a source map, source map consumers can use the resulting Decoded Source Map Records to look up position information for debugging or other use cases. This section describes the behaviour of typical operations that may be supported by source map consumers.</p>
+  <p>The GetOriginalPosition operation may be used to query the position in an original source that corresponds to a position in the generated code. For example, this can be used in a debugger to navigate from the generated code to the original source based on a user's mouse click.</p>
+
+  <emu-clause id="sec-GetOriginalPosition" type="abstract operation">
+    <h1>
+      GetOriginalPosition (
+        _sourceMapRecord_: a Decoded Source Map Record,
+        _generatedPosition_: a Position Record,
+      ): an Original Position Record or *null*
```
takikawa, line 1553: I also changed this to return the original position record instead of the entry. Given the name of the operation it probably makes more sense to return the position record, and that also makes a bit more sense for the range mappings proposal spec text (since you just return a modified position, rather than making up a new mapping with a modified position).

This should also be ok for PR #219 which only uses the original position part.

nicolo-ribaudo (approved):

spec.emu
```diff
@@ -1545,6 +1545,37 @@ return lastURL;</code></pre>
...
+      ): a List of Original Position Records
+    </h1>
+    <dl class="header"></dl>
+    <emu-alg>
+      1. Let _mappings_ be _sourceMapRecord_.[[Mappings]].
+      1. Let _last_ be *null*.
+      1. Let _originalPositions_ be a new empty List.
+      1. For each element _mapping_ of _mappings_, in reverse List order, do
+        1. If the result of performing ComparePositions(_mapping_.[[GeneratedPosition]], _generatedPosition_) is ~lesser~ or ~equal~, then
+          1. If _last_ is *null*, then
```
nicolo-ribaudo, line 1568: Nit: swap these two conditions, to make it clear that we basically have an early exit from the loop.

spec.emu
```diff
@@ -1545,6 +1545,37 @@ return lastURL;</code></pre>
...
+        _sourceMapRecord_: a Decoded Source Map Record,
+        _generatedPosition_: a Position Record,
+      ): a List of Original Position Records
+    </h1>
+    <dl class="header"></dl>
+    <emu-alg>
+      1. Let _mappings_ be _sourceMapRecord_.[[Mappings]].
+      1. Let _last_ be *null*.
+      1. Let _originalPositions_ be a new empty List.
+      1. For each element _mapping_ of _mappings_, in reverse List order, do
```
jridgewell, line 1566: Nit: why not do a forward scan, instead of a reverse _and_ forward scan?

```js
let matches = [];
for (const mapping of mappings) {
  const compare = comparePos(mapping, needle);
  if (compare === EQUAL) {
    matches.push(mapping);
  } else if (compare === GREATER) {
    break;
  }
}
```

jridgewell (approved):

spec.emu
takikawa: As it's currently written, the function returns all the mappings at the equal generated position if there are any, or all the mappings that's at the greatest lower bound generated position if there aren't equal mappings.

If you only do a forward scan you find the equal mappings easily but you have to do more work for the lower bound.

PR #198, opened 2025-04-11
fix: Improve 'emu-note' text contrast w/ dark mode

rschristian:
Now that ecmarkup supports dark mode, the emu-notes cause a bit of a readability issue (in that they're impossible to read) as they set light backgrounds which contrast poorly with the page-wide foreground color. Setting the foreground color to black for both seems appropriate.

Image, for reference:

Shows the current styling of example & issue notes, using white text on a very light background so that text is impossible to read

> https://tc39.es/ecma426/#sec-linking-inline

The hljs setup will need to be adjusted yet too, though that can be done separately by someone else. Seems a bit more involved and would need some playing with, whilst this is just a drive-by fix.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -37,6 +37,7 @@
   }
 
   emu-note[issue], emu-note[example] {
+    color: black;
     counter-increment: var(--note-type);
     border-left-color: var(--note-border-color);
     background-color: var(--note-background-color);


rschristian:
The contributing guide doesn't differentiate between spec changes and everything else, hopefully this is fine? If not, certainly feel free to copy this under someone else's name if need be.

nicolo-ribaudo:
I _think_ this is not subject to IPR, because you cannot reasonably claim any IPR on setting `color: black` in this, but just in case would you mind filling https://tc39.es/agreements/contributor/?

nicolo-ribaudo:
Actually wait, let me double-check that it's the right form.

nicolo-ribaudo:
Yes ok it's the correct one.

nicolo-ribaudo:
Well it seems like CI liked it. Thank you! :)

nicolo-ribaudo (approved):

PR #199, opened 2025-04-15
Add Bugsink as a "consumer" to debug-id.md

vanschelven:


vanschelven:
I did read the contributing guidelines, and realize that those state that to contribute, I'd need to join the TC;

happy to do that if indeed required; for now I'm just leaving this here "out of order", as I imagine that the single act of documenting the existence of a consumer of debug IDs may not have such strict requirements, or otherwise, someone who _is_ on the TC is willing to "sponsor" this PR.

vanschelven:
> simply re-use Sentry code [..] symbolicator [..] Sentry backend [..] Sentry tools

I get the point; let's leave the weighing of that point to other TC39 members. In the meantime I'll answer.

* for the _injection_ of Debug IDs, Bugsink relies entirely on other (Sentry) tools, in particular on `sentry-cli`
* for the _consumption_ of the Debug IDs, there is not external dependency at all, Bugsink simply has its own implementation of this functionality. Given the header "Consumers/Debuggers" I'd argue this is the relevant fact :smile: 

for completeness: for sourcemaps parsing, Bugsink does not rely on symbolicator, although it currently does rely on another [Open Source library](https://github.com/mattrobenolt/python-sourcemap)

lforst (approved):

nicolo-ribaudo:
I was also a little bit conflicted about this, for a different reason: I don't want proposals to just become a list of "advertisement" for products. It's good however to show that there are multiple real-world use cases for the proposal, and that it's not just Sentry doing their own thing.

That said, a list containing two links is far from being what I consider problematic :P

I'll merge this pr as @lforst approved it. Luca, I'll give you write access to the repo so that you can merge PRs related to DebugID.

PR #200, opened 2025-04-15
CONTRIBUTING: match quantifier with actual number of items in the list

vanschelven:


vanschelven:
I fully understand that this is an entirely trivial change, but you "gotta start somewhere"; 

alternatively, I can imagine that there's actually a third group missing from the enumeration below, or that the enumeration should be split up differently. In that case, just close my PR

jridgewell (approved):

nicolo-ribaudo:
@vanschelven Thank you for your PR! Could you fill https://tc39.es/agreements/contributor/?

vanschelven:
I just did

Klaas van Schelven
+31 6 811 599 10

Issue #146, opened 2024-10-30
Debug IDs Proposal: Should We Define a JavaScript API?


lforst:
This issue relates to the [Debug ID proposal](https://github.com/tc39/source-map/blob/main/proposals/debug-id.md).

While we’ve made good progress in defining how Debug IDs should be represented in generated artifacts through the `debugId` field and `//# debugId=...` comments, we are still slightly blocked on the part of the proposal that suggests how engines and languages should interact with Debug IDs.

When superficially looking at implementing a JS API in various engines, a fundamental question came up: **Should we even propose and specify a runtime API?**

### Considerations

1. **Scope of the Source Map Spec**  
   First up, there is the issue of the Source Map Spec technically being entirely separate from the ECMAScript and HTML Specs. It feels somewhat nonsensical to put a specification for a runtime API into the Source Map Spec, especially since the Spec is mostly language agnostic.

2. **Challenges in API Design**
   When exploring potential designs for a JavaScript API, several complications arise:

   - **Approach 1: `getDebugIdForUrl(url: string): string | null;`**  
     A straightforward idea is to create an API like `getDebugIdForUrl()`, which, given a file URL, retrieves the corresponding Debug ID. For example, upon encountering an error, we could parse the stack trace, extract each frame’s file URL, and pass it to `getDebugIdForUrl()` to obtain its Debug ID.

     However, this approach has a critical flaw: the `Error.stack` property, which would be necessary for parsing the stack trace, is not formally specified. This makes it unreliable as a foundation for a standardized API, since it relies on non-standardized conventions across engines.

   - **Approach 2: `Error.getDebugIdsForError(err: Error): (string | null)[];`**  
     Having established that relying on `Error.stack` is not great, an alternative would be a static API `Error.getDebugIdsForError()`, which, when passed an error object, returns an array of Debug IDs for each stack frame, with `null` indicating a missing Debug ID. This approach is a bit more functional it abstracts away the specifics of stack trace parsing and Debug ID retrieval.

     However, this API also comes with engine-specific challenges. In V8 it seems doable with a reasonable and lazy evaluation. In JavaScriptCore it becomes a bit harder because the stack is captured at Exception-Time and we would probably have to grab the Debug IDs at that point in time too. For Spidermonkey it will likely work out similar to JSC. Because for 2 out of 3 engines, this implementation would have a non-trivial runtime impact, lending itself to an opt-in situation (`Error.captureDebugIds = true`), it may again be hard to find a standardized specification.

[...]

PR #201, opened 2025-04-15
Move JS API out of debug ID proposal

lforst:
- No longer suggest a JS API in the main part of the proposal (we still discuss possible options in an appendix to highlight the value for post-hoc debugging)
- Be more specific about JS APIs (ref https://github.com/tc39/ecma426/issues/146)
- Remove unnecessary details from symbol server section (we don't care about the storage mechanism)

cc @szuend

### DIFF
--- a/proposals/debug-id.md
+++ b/proposals/debug-id.md
@@ -36,7 +36,7 @@ The proposed solution offers the following benefits:
 
 5. Guaranteed bidirectionality: Today source maps do not provide the ability to reliably resolve back to the generated file they are from. However in practice tools often require this as they are often leveraging the generated artifact to resolve scope information by parsing the source.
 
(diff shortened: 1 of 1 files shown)


szuend (approved):
Thanks!

I think it's also worth to call out an additional use-case for debugIDs, which is source map caching in debuggers. Source maps keyed on URLs can't be cached without loading them + hashing the content. With debugIDs, browser dev tools and IDEs can decide upfront whether they need to load + parse a source map or not.

lforst:
Great point. Thank you! I added it to the use-cases section.

I also added a short explainer that whether or not a debug ID is deterministic is encoded in the version bits of the UUID, which is an important bit (pun not intended) of information since devtools and debuggers may probably only want to apply caching mechanisms when the debug ID is deterministic.

lforst:
@szuend @nicolo-ribaudo I think this would be fine to merge if any of you want to do it. :) I don't have merge permissions myself.

PR #205, opened 2025-04-25
Scopes: Use unsigned VLQ (1-based) for binding expressions

szuend:
As discussed in the meeting of 2025-04-24, this PR changes the binding expressions from a signed VLQ to a unsigned VLQ. To signify a variable is unavailable in a given generated (sub) range, we use `0` instead of `-1` and the indices themselves are now 1-based instead of 0-based. This is necessary to free up `0` as the "unavailability marker".

To capture some of the points of the meeting here (also available in the notes):

- Using relative indicies into the `names` array would require us to sacrifice an additional bit to signify "unavailability".
- Another option would be to use `0` as the marker and make positive relative indices "off by one". E.g. "0" means a variable is unavailable, "1" means a variable uses the same binding expression as the previous variable and "2" means it uses the next entry in "names".
- Both of the above options show some minor improvement in the uncompressed size of the scopes encoding but are actually worse after gzip/brotli compression.
- This leads us to this PR as to not overcomplicate the encoding for no gain.

### DIFF
--- a/proposals/scopes.md
+++ b/proposals/scopes.md
@@ -363,12 +363,12 @@ If a "generated range" contains a callsite, then the range describes an inlined 
 ```
 generated_range_bindings :=
   'G'       // Tag: 0x6 unsigned
-  sBINDING+
+  uBINDING+
 ```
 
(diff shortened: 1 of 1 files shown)


szuend:
@nicolo-ribaudo adding you explicitly since you didn't have a chance yet to express your opinion. I remember you floated the idea of using `B` (the minimum possible integer) to mark "unavailability" of a variables' value. We didn't discuss that option though in yesterdays' meeting.

hbenl (approved):

PR #211, opened 2025-06-11
Avoid ambiguity in regexp-based extraction

nicolo-ribaudo:
This PR changes:
- the non-regexp-based logic to only allow single-line comment (matching browser implementations)
- the regexp-based approach to start from the end and bail out whenever there is either code or a comment containing `` ` ``, `'`, `"`, or `*/`.

The notes still need to be updated.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1328,7 +1328,7 @@
         <p>The following JavaScript code links to a source map, but it does not do so <emu-xref href="#umambiguous-linking">unambiguously</emu-xref>:</p>
 
         <pre><code class="javascript">let a = &#x60;
-// sourceMappingURL=foo.js.map
+//# sourceMappingURL=foo.js.map
 // &#x60;</code></pre>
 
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -1350,7 +1350,7 @@
         <emu-alg>
           1. Let _tokens_ be the List of tokens obtained by parsing _source_ according to <emu-xref href="#sec-ecmascript-language-lexical-grammar">ECMA-262's lexical grammar</emu-xref>.
           1. For each nonterminal _token_ in _tokens_, in reverse order, do
-            1. If _token_ is not |SingleLineComment| or |MultiLineComment|, return *null*.
+            1. If _token_ is not |SingleLineComment|, return *null*.
```
sokra, line 1353: Also check for disallowed chars here so that both implementation match up

jkup:
Some context (cc @DanielRosenwasser)

When we were first exploring the `//#sourceMappingUrl` concept last year we proposed adding to the official specification that you *must* parse the file, find the last comment and check it's URL. We quickly got feedback from the VSCode team that this approach would be too slow for them and they have a need to quickly extract the comment URL without parsing the entire source code.

So when we presented the specification to tc39 last year we pitched offering both approaches. We got some pushback that the regex approach should have the same functionality as the parsing approach (never returning different results) or at least that one approach should be a subset of the other. This PR is @nicolo-ribaudo's attempt at updating the regex approach to align with that goal but we wanted to make sure that this approach still works well for TypeScript and VSCode.

takikawa (approved):
LGTM modulo the two comments.

PR #215, opened 2025-06-26
Try fixing IPR

nicolo-ribaudo:
Fyi @lforst all the commits on GitHub that you did with your Sentry email are not linked to your account anymore (example: https://github.com/tc39/ecma426/commit/063d1a93a1e8610680a0fc9e0e1d3600ae4ef185). You might want to keep the email in your GH settings even if it doesn't exist anymore (as a secondary email) so that they keep linking to your account.


nicolo-ribaudo:
It runs the job from `main`, so the only way to test it is merging.

PR #219, opened 2025-08-12
Scopes: Algorithm for computing original stack frames

hbenl:
## [Preview](https://tc39.es/ecma426/pr/219/)

This is an initial draft of the algorithm for computing original stack frames (implemented [here](https://github.com/hbenl/tc39-proposal-scope-mapping/blob/master/src/getOriginalFrames.ts)).
This PR makes use of `GetOriginalPosition` from #195.

TODO:
- ~handle multiple generated sources and sourcemaps~
- ~handle outlined functions~
- handle frames with generated locations that have no mapping

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -2638,6 +2638,413 @@
   </emu-clause>
 </emu-clause>
 
+<emu-clause id="sec-computing-original-stack-frames">
+  <h1>Computing original stack frames</h1>
+  <p></p>
+
+  <emu-clause id="sec-source-maps-record-type">
+    <h1>Source Maps Record</h1>
+    <p>A <dfn id="source-maps-record">Source Maps Record</dfn> associates source URLs to Decoded Source Map Records.</p>
+  </emu-clause>
+
+  <emu-clause id="sec-debugger-position-record-type">
+    <h1>Debugger Position Record</h1>
+    <p>A <dfn id="debugger-position-record">Debugger Position Record</dfn> is a tuple of a source URL and non-negative line and column numbers.</p>
+    <emu-table id="table-debugger-position-record-fields">
+      <table>
+        <thead>
+          <tr>
+            <th>Field Name</th>
+            <th>Value Type</th>
+          </tr>
+        </thead>
+        <tr>
+          <td>[[URL]]</td>
+          <td>a URL or *null*</td>
+        </tr>
+        <tr>
+          <td>[[Line]]</td>
+          <td>a non-negative integral Number</td>
+        </tr>
+        <tr>
+          <td>[[Column]]</td>
+          <td>a non-negative integral Number</td>
+        </tr>
+      </table>
+    </emu-table>
+  </emu-clause>
+
+  <emu-clause id="sec-debugger-bindings-record-type">
+    <h1>Debugger Bindings Record</h1>
+    <p>A <dfn id="debugger-bindings-record">Debugger Bindings Record</dfn> associates variable names to implementation-defined representations of javascript values in the debuggee or to ~unavailable~.</p>
+  </emu-clause>
+
+  <emu-clause id="sec-debugger-scope-record-type">
+    <h1>Debugger Scope Record</h1>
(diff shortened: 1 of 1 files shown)


szuend:
Thanks for writing up the algorithm! I had a cursory initial look.

IMHO `ComputeOriginalFrames` does potentially too much: It'll also have to handle inlining/outlining of frames. Together with building the scope chain for every frame such an abstract operation does a lot.

Rather then having one catch all abstract operation I suggest we spec out some basic building blocks. Concretely I suggest the following 3 "entry point operations":

* `FindOriginalFunctionName(position: GeneratedPosition): string|null`

  This is already somewhat in the current PR, albeit spread out. I'd spec it as:
  
  1. Find inner-most generated range for `position` (already there).
  1. Take the `definition` scope of the inner-most generated range
  1. Walk it outwards until we find a scope with `isStackFrame = true`.
  
* `SymbolizeStackTrace(rawFrames: Array<{url: URL, position: GeneratedPosition}>): Array<{name?: string, position: OriginalPosition}>`

  Doesn't handle scopes at all, purely translates a stack trace from generated source to original source. Note that `rawFrames` can contain frames across different bundles so we need some "implementation-defined" operation to get at the scope information via a raw frames URL.

  This operation expands inlined frames, collapses outlined frames and utilizes `FindOriginalFunctionName` and `GetOriginalPosition` to translate frames.

* `BuildScopeChain(position: GeneratedPosition): OriginalScopeWithValues[]`

  Where `OrigianlScopeWithValues` is a record with:
    * The `OriginalScope`
    * A map of original variable name to the concrete JS value

  This operation does part of whats in `ComputeOriginalFrame` of the PR: Find the inner-most range and get it's definition scope. Walk that scope outwards and also find the corresponding generated range for each scope in the chain.

  I'd suggest though that we don't spec out how the evaluation works: Both Firefox and Chrome have the ability to evaluate expressions in the lexcial environment of a paused position but there might be restrictions. As such, I think it would ok to require a host-defined operation `EvaluateInRange(range: GeneratedRange, expression: string): JSValue`.

  We can leave it up to debuggers how to find the right debugger scope to evaluate `expression` in. We'll already spec how generated ranges relate to concrete ECMAScript grammar productions, so implementations should be able to then find the right debugger scope from that.

szuend:
* Good point that `FindOriginalFunctionName` can return more than one name. Note though that it's kind of the special case of my `SymbolizeStackTrace` with only a single frame.
* Also good point that `BuildScopeChain` needs an original position. Alternatively we could pass in an "inlined frame index". The number of inlined frames is basically the length of the result of `SymbolizeStackTrace` with a single generated position.
* FWIW, Chrome doesn't have the possibility to pass custom bindings via protocol and this would also be non-trivial to implement. On the other hand, I have concrete plans to add a `scopeIndex` to the `Debugger.evaluateOnCallFrame` command, so expressions can be evaluated in an arbitrary scope of a call frame. That is why I'd prefer to keep this implementation-defined. Different engines handle "debug evaluate" or "evaluate in a lexical environment" differently and offer different capabilities, so I don't want to prescribe requirements that no or only half the implementation end up following.

hbenl:
> I have concrete plans to add a scopeIndex to the Debugger.evaluateOnCallFrame command, so expressions can be evaluated in an arbitrary scope of a call frame.

Nice!

> I don't want to prescribe requirements that no or only half the implementation end up following

OK, we could use "evaluate in this scope" in the algorithm and then append a description how "evaluate in this scope" can be emulated using "evaluate with bindings".

szuend:
> > Chrome doesn't have the possibility to pass custom bindings via protocol
> 
> I think this can be emulated using [Runtime.callFunctionOn](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn).

I'm not sure how'd you create the bindings object in the first place. You'd still need to read the values for generated variables off the stack or out of closure contexts. [Debugger.evaluateOnCallFrame](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-evaluateOnCallFrame) has this built-in by basically building `with` scopes for each scope that know how to read from the stack or closure contexts.

At least from a V8 and Chrome DevTools perspective it would be much easier to just extend `evaluateOnCallFrame` to allow expression evaluation in an arbitrary scope returned by the `Debugger.paused` event.

PR #220, opened 2025-08-28
Scopes: Refactor scopes tree decoding

nicolo-ribaudo:
Originally opened at https://github.com/szuend/source-map/pull/1. See https://github.com/tc39/ecma426/pull/196#issuecomment-3211293112.

I tried to make it a bit easier to see what is going on by:

- giving different names to the different SDOs that take care of parsing different data about the scopes
- getting all the data and then creating the Original Scope Record, rather than creating an Original Scope Record and then mutating it
- centralizing the "increment the index in the state and get the result" in a single AO

We could do something similar for TopLevelItemList.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1775,59 +1775,58 @@
         </emu-clause>
       </emu-clause>
 
-        <emu-clause id="sec-DecodedOriginalScopeTrees" type="sdo">
-          <h1>
-            DecodedOriginalScopeTrees (
-              _state_: a Decode Scope State Record,
-              _names_: a List of Strings,
-            ): a List of either Original Scope Records or *null*
-          </h1>
-          <dl class="header"></dl>
-          <emu-grammar>
-            OriginalScopeItemList : OriginalScopeItemList `,` OriginalScopeItem
-          </emu-grammar>
-          <emu-alg>
-            1. Let _left_ be the DecodedOriginalScopeTrees of |OriginalScopeItemList| with arguments _state_ and _names_.
-            1. Let _right_ be the DecodedOriginalScopeTrees of |OriginalScopeItem| with arguments _state_ and _names_.
-            1. Return the list-concatenation of _left_ and _right_.
-          </emu-alg>
-          <emu-grammar>
-            OriginalScopeTreeItem: [empty]
-          </emu-grammar>
-          <emu-alg>
-            1. Return « *null* ».
-          </emu-alg>
-          <emu-grammar>
-            OriginalScopeTree :
-              OriginalScopeStart OriginalScopeVariablesItem? OriginalScopeItemList? `,` OriginalScopeEnd
-          </emu-grammar>
-          <emu-alg>
-            1. Let _start_ be the OriginalScopePosition of |OriginalScopeStart| with arguments _state_.[[ScopeLine]] and _state_.[[ScopeColumn]].
-            1. Let _name_ be the OriginalScopeName of |OriginalScopeStart| with arguments _state_.[[ScopeNameIndex]] and _names_.
-            1. Let _kind_ be the OriginalScopeKind of |OriginalScopeStart| with arguments _state_.[[ScopeKindIndex]] and _names_.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
I just noticed that this makes FlattenedScopes in post-order rather than pre-order, I need to fix that.

spec.emu
```diff
@@ -1755,159 +1775,170 @@
...
+            <dt>description</dt>
+            <dd>It returns the name referenced by _vlq_, considering it as a relative index in _names_.</dd>
+          </dl>
           <emu-alg>
-            1. Let _relativeKind_ be the VLQSignedValue of _kind_.
-            1. Set _state_.[[ScopeKindIndex]] to _state_.[[ScopeKindIndex]] + _relativeKind_.
-            1. Set _scope_.[[Kind]] to _names_[_state_.[[ScopeKindIndex]]].
+            1. Let _relativeIndex_ be the VLQSignedValue of _vlq_.
+            1. Let _index_ be AccumulateIndex(_accumulator_, _relativeIndex_).
+            1. If _index_ > the length of _names_, then
```
szuend, line 1938:
```suggestion
            1. If _index_ >= the length of _names_, then
```

Off-by-one?

spec.emu
```diff
@@ -1705,36 +1741,20 @@
...
+            1. Let _originalScopes_ be the DecodedOriginalScopeTrees of |OriginalScopeTreeList| with arguments _state_ and _names_.
+            1. Set _info_.[[Scopes]] to _originalScopes_.
           </emu-alg>
           <emu-grammar>
-            OriginalScopeTreeItem :
-              OriginalScopeTree
-          </emu-grammar>
-          <emu-alg>
-            1. Set _state_.[[ScopeLine]] to 0.
-            1. Set _state_.[[ScopeColumn]] to 0.
```
szuend, line 1728: Did this get ported over? I was not able to find the resetting of the scope line/column state in-between top-level original scopes.

szuend (changes requested):

spec.emu
```diff
@@ -1755,159 +1775,189 @@
...
+            1. Let _name_ be the OriginalScopeName of |OriginalScopeStart| with arguments _state_.[[ScopeNameIndex]] and _names_.
+            1. Let _kind_ be the OriginalScopeKind of |OriginalScopeStart| with arguments _state_.[[ScopeKindIndex]] and _names_.
+            1. Let _flags_ be the VLQUnsignedValue of |OriginalScopeStart|'s |ScopeFlags| nonterminal.
+            1. Let _originalScope_ be the Original Scope Record { [[Start]]: _start_, [[End]]: _start_, [[Name]]: _name_, [[Kind]]: _kind_, [[Variables]]: « », [[Children]]: « », [[IsStackFrame]]: *false* }.
+            1. Append _originalScope_ to _state_.[[FlatOriginalScopes]].
+            1. If _flags_ & 0x4 = 0x4, set _originalScope_.[[IsStackFrame]] to *true*.
+            1. If |OriginalScopeVariablesItem| is present, then
+              1. Set _originalScope_.[[Variables]] to the OriginalScopeVariables of |OriginalScopeVariablesItem| with arguments _state_.[[ScopeVariableIndex]] and _names_.
+            1. If |OriginalScopeItemList| is present, then
+              1. Set _originalScope_.[[Children]] to the DecodedOriginalScopeTrees of |OriginalScopeItemList| with arguments _state_ and _names_.
```
szuend, line 1832: Should this be `DecodeInnerOriginalScopeTrees`?

szuend (approved):

szuend:
Thanks! lgtm % a minor bug.

As a suggestion: We could flip the naming. So instead of having `DecodeOriginalScopeTree` and `DecodeInnerOriginalScopeTree`, we could do `DecodeTopLevelOriginalScopeTree` and `DecodeOriginalScopeTree`. It fits slightly better since we "execute" `DecodeInnerOriginalScopeTree` even for a top-level scope.

PR #227, opened 2025-09-10
Scopes: Introduce 'Position Accumulator Record'

szuend:
The new 'AccumulatePosition' will be  re-used for generated range position and sub-range binding positions. That way, we don't have to duplicate the "reset column when line is non-zero" logic everywhere.

Drive-by: Rename "OriginalScopePosition" to "DecodedPosition" so we may use the same SDO also for the GeneratedRangeStart and GeneratedRangeEnd productions.


nicolo-ribaudo (approved):
We might also want to already use these in the main spec for mappings.

PR #231, opened 2025-10-02
Scopes: Refactor generated range tree decoding

szuend:
This PR does the equivalent for #220 but for generated ranges.

For sub-range bindings there are two different approaches we can take:
- Pass `_bindings_` into a sub-range SDO and update `_bindings_[_index_]` as it is decoded.
- Add a new sub-range binding record ( `{ index: number, bindings: BindingRecord[] }`) and add an SDO that returns an array of these sub-range binding records of all the immediate `|GeneratedSubRangeBinding|` children. These are then later combined with `_bindings_`.

In the spirit of not modifying state, this PR chooses the latter but I don't mind changing it to the former if that is preferred.

Drive-bys:
* Rename `|RangeBindingList|` to `|BindingExpressionList|` and unify `|RangeBinding|` with `|BindingExpression|`.
* Fix linking for binding records and named mappings.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1091,7 +1091,7 @@
         <li>The original source language construct maps semantically to the generated JavaScript code.</li>
         <li>The original source language construct has a name.</li>
       </ul>
-      <p>Then the [[Name]] of the <emu-xref href="#decoded-mapping-record">mapping</emu-xref> entry should be the name of the original source language construct. A <emu-xref href="#decoded-mapping-record">mapping</emu-xref> with a non-null [[Name]] is called a <dfn variant="named mappings">named mapping</dfn>.</p>
+      <p>Then the [[Name]] of the <emu-xref href="#decoded-mapping-record">mapping</emu-xref> entry should be the name of the original source language construct. A <emu-xref href="#decoded-mapping-record">mapping</emu-xref> with a non-null [[Name]] is called a <dfn variants="named mappings">named mapping</dfn>.</p>
 
       <emu-note example>
         A minifier renaming functions and variables or removing function names from immediately invoked function expressions.
@@ -1288,7 +1288,7 @@
 
       <emu-clause id="sec-binding-record-type">
         <h1>The Binding Record</h1>
-        <p>The <dfn variant="Binding Records">Binding Record</dfn> describes a JavaScript expression that must be used for the given position to retrieve a specific variables' value.</p>
+        <p>The <dfn variants="Binding Records">Binding Record</dfn> describes a JavaScript expression that must be used for the given position to retrieve a specific variables' value.</p>
         <emu-table id="table-binding-record-fields" caption="Binding Record Fields">
           <table>
             <thead>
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -2008,107 +1982,90 @@
...
+        </emu-alg>
+        <emu-grammar>
+          GeneratedRangeTree :
+            GeneratedRangeStart GeneratedRangeBindingsItem? GeneratedRangeCallSiteItem? GeneratedRangeItemList? `,` GeneratedRangeEnd
+        </emu-grammar>
+        <emu-alg>
+          1. Let _start_ be the DecodedPosition of |GeneratedRangeStart| with argument _state_.[[RangePosition]].
+          1. Let _definition_ be the GeneratedRangeDefinition of |GeneratedRangeStart| with arguments _state_.[[RangeDefinition]] and _state_.[[FlatOriginalScopes]].
+          1. Let _flags_ be the VLQUnsignedValue of |GeneratedRangeStart|'s |RangeFlags| nonterminal.
+          1. If _flags_ & 0xc = 0xc, let _stackFrameType_ be ~hidden~.
```
takikawa: In the previous version, I think this would optionally error if `0x8` was on but `0x4` was off, while this version sets it to the NONE type. Was that an intentional change?

spec.emu
```diff
@@ -2117,83 +2074,138 @@
...
           <emu-alg>
-            1. Perform DecodeGeneratedRangeItem of |BindingLine| with arguments _range_, _state_ and _names_.
-            1. Perform DecodeGeneratedRangeItem of |BindingColumn| with arguments _range_, _state_ and _names_.
-            1. Let _bindingIndex_ be the VLQUnsignedValue of |BindingExpression|.
-            1. If _bindingIndex_ > 0, let _binding_ be _names_[_bindingIndex_ - 1].
-            1. Else, let _binding_ be *null*.
-            1. Append { [[From]]: { [[Line]]: _state_.[[SubRangeLine]], [[Column]]: _state_.[[SubRangeColumn]] }, [[Binding]]: _binding_ } to _range_.[[Bindings]][_state_.[[SubRangeVariableIndex]]].
+            1. Let _variableIndex_ be the VLQUnsignedValue of |VariableIndex|.
+            1. Let _from_ be a new Position Accumulator Record { [[Line]]: _start_.[[Line]], [[Column]]: _start_.[[Column]] }.
+            1. Let _bindings_ be the SubRangeBindingList of |BindingFromList| with arguments _from_ and _names_.
```
takikawa: I think there's a typo here for the SDO: `SubRangeBindingList` -> `SubRangeBinding`

spec.emu
```diff
@@ -2117,83 +2074,138 @@
...
+            1. If _unadjustedBindingIndex_ = 0, return *null*.
+            1. Let _bindingIndex_ be _unadjustedBindingIndex_ - 1.
+            1. If _bindingIndex_ >= the length of _names_, then
+              1. Optionally report an error.
+              1. Return *null*.
+            1. Return _names_[_bindingIndex_].
           </emu-alg>
+        </emu-clause>
+
+        <p>A <dfn variants="Sub-Range Binding Records">Sub-Range Binding Record</dfn> is the result of decoding a |GeneratedSubRangeBinding|. It has the following fields:</p>
```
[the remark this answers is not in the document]
szuend: Done

PR #232, opened 2025-10-08
Range mappings: add draft spec language to branch

takikawa:
This PR adds the draft range mappings spec language to the proposal branch. Once this is merged I'll open a separate PR from the proposal branch to main for detailed comments and review.

PR #233, opened 2025-10-20
Range mappings: spec text for proposal

takikawa:
This is the actual PR for the draft range mappings proposal for detailed review. Comments and feedback welcome.

On the other PR #232, Nic raised some points about what we should do in case there are multiple mappings in one location and only one (or a subset) are range mappings.

https://tc39.es/ecma426/pr/233/

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -684,9 +684,10 @@
         1. Let _ignoreListField_ be GetOptionalListOfArrayIndexes(_json_, *"ignoreList"*).
         1. Let _sources_ be DecodeSourceMapSources(_baseURL_, _sourceRootField_, _sourcesField_, _sourcesContentField_, _ignoreListField_).
         1. Let _namesField_ be GetOptionalListOfStrings(_json_, *"names"*).
+        1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_).
         1. <ins>Let _rangeMappingsField_ be GetOptionalString(_json_, *"rangeMappings"*).</ins>
         1. <ins>Let _rangeMappings_ be DecodeRangeMappings(_rangeMappingsField_).</ins>
-        1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_, <ins>_rangeMappings_</ins>).
+        1. <ins>Perform UpdateMappingsForRangeMappings(_mappings_, _rangeMappings_).</ins>
         1. [declared="a,b"] Sort _mappings_ in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if ComparePositions(_a_.[[GeneratedPosition]], _b_.[[GeneratedPosition]]) is ~lesser~.
         1. Return the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_ }.
       </emu-alg>
@@ -921,10 +922,6 @@
             <td>[[NameIndex]]</td>
             <td>a non-negative integer</td>
           </tr>
-          <tr>
-            <td><ins>[[MappingIndex]]</ins></td>
-            <td><ins>a non-negative integer</ins></td>
-          </tr>
         </table>
       </emu-table>
 
@@ -935,7 +932,6 @@
             _mappings_: a List of Decoded Mapping Records,
             _names_: a List of Strings,
             _sources_: a List of Decoded Source Records,
-            <ins>_rangeMappings_: a List of of Decoded Range Mapping Offset Records,</ins>
           )
         </h1>
         <dl class="header"></dl>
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -947,47 +962,50 @@
...
           1. Else,
             1. Let _originalPosition_ be a new Original Position Record { [[Source]]: _sources_[_state_.[[SourceIndex]]], [[Line]]: _state_.[[OriginalLine]], [[Column]]: _state_.[[OriginalColumn]] }.
           1. Let _name_ be *null*.
           1. If |Name| is present, then
-            1. Perform DecodeMappingsField of |Name| with arguments _state_, _mappings_, _names_ and _sources_.
+            1. Perform DecodeMappingsField of |Name| with arguments _state_, _mappings_, _names_<ins>,</ins> <del>and</del> _sources_ <ins>and _rangeMappings_</ins>.
             1. If _state_.[[NameIndex]] &lt; 0 or _state_.[[NameIndex]] ≥ the number of elements of _names_, optionally report an error.
             1. Else, set _name_ to _names_[_state_.[[NameIndex]]].
-          1. Let _decodedMapping_ be a new DecodedMappingRecord { [[GeneratedPosition]]: _generatedPosition_, [[OriginalPosition]]: _originalPosition_, [[Name]]: _name_ }.
+          1. <ins>Let _isRangeMapping_ be the result of performing LookupRangeMapping with arguments _rangeMappings_, _state.[[GeneratedLine]], and _state.[[MappingIndex]].</ins>
```
nicolo-ribaudo:
```suggestion
          1. <ins>Let _isRangeMapping_ be the result of performing LookupRangeMapping with arguments _rangeMappings_, _state_.[[GeneratedLine]], and _state_.[[MappingIndex]].</ins>
```

spec.emu
```diff
@@ -682,7 +684,10 @@
         1. Let _ignoreListField_ be GetOptionalListOfArrayIndexes(_json_, *"ignoreList"*).
         1. Let _sources_ be DecodeSourceMapSources(_baseURL_, _sourceRootField_, _sourcesField_, _sourcesContentField_, _ignoreListField_).
         1. Let _namesField_ be GetOptionalListOfStrings(_json_, *"names"*).
-        1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_).
+        1. <ins>Let _rangeMappingsField_ be GetOptionalString(_json_, *"rangeMappings"*).</ins>
+        1. <ins>If _rangeMappingsField_ is *null*, set _rangeMappingsField_ to *""*.</ins>
```
jridgewell: Nit: just handle this in the decode function?

spec.emu
```diff
@@ -1119,6 +1138,142 @@
...
+            1. Perform DecodeRangeMappingsField of |RangeMappingList| with arguments _state_, _rangeMappings_.
+          </emu-alg>
+          <emu-grammar>
+            RangeMapping :
+              Vlq
+          </emu-grammar>
+          <emu-alg>
+            1. Let _relativeIndex_ be the VLQUnsignedValue of |Vlq|.
+            1. If _relativeIndex_ is 0, then
+              1. Optionally report an error.
```
jridgewell: Does this mean the first mapping can never be a range mapping?

```js
"mappings": "AAAA",
"rangeMappings": "A"
```

spec.emu
```diff
@@ -1119,6 +1138,142 @@
...
+            1. Set _state_.[[MappingIndex]] to _state_.[[MappingIndex]] + _relativeIndex_.
+            1. Let _rangeMappingOffset_ be a new Decoded Range Mapping Offset Record { [[GeneratedLine]]: _state_.[[GeneratedLine]], [[MappingIndex]]: _state_.[[MappingIndex]] - 1 }.
```
jridgewell: Why are we dealing with 1-based indexes?

spec.emu
```diff
@@ -931,10 +945,11 @@
             Line `;` LineList
         </emu-grammar>
         <emu-alg>
-          1. Perform DecodeMappingsField of |Line| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. Perform DecodeMappingsField of |Line| with arguments _state_, _mappings_, _names_<ins>,</ins> <del>and</del> _sources_ <ins>and _rangeMappings_</ins>.
           1. Set _state_.[[GeneratedLine]] to _state_.[[GeneratedLine]] + 1.
           1. Set _state_.[[GeneratedColumn]] to 0.
-          1. Perform DecodeMappingsField of |LineList| with arguments _state_, _mappings_, _names_ and _sources_.
+          1. <ins>Set _state_.[[MappingIndex]] to 0.</ins>
+          1. Perform DecodeMappingsField of |LineList| with arguments _state_, _mappings_, _names_<ins>,</ins> <del>and</del> _sources_ <ins>and _rangeMappings_</ins>.
```
jridgewell: I think we should leave `DecodeMappingsField` alone, and instead pass `mappings` to `DecodeRangeMappings`. It allows us define error cases for 
- range mapping point to a `GenColumn` sourceless mapping
- range mapping points to undefined mapping (line too large or index too large)

spec.emu
```diff
@@ -1119,6 +1128,168 @@
...
+        <emu-clause id="sec-FindMappingForRangeMappingOffset" type="abstract operation">
+          <h1>
+            FindMappingForRangeMappingOffset (
+              _mappings_: a List of Decoded Mapping Records,
+              _offset_: a Decoded Range Mapping Offset Record,
+            ): a Decoded Mapping Record or ~not-found~
+          </h1>
+          <dl class="header"></dl>
+          <emu-alg>
+            1. Let _currentLine_ be *null*.
```
jridgewell: Nit: `-1`, so we have a consistent number type?

spec.emu
jridgewell: Going though this in code, I still hate it. When I'm reading the relative offsets, I don't want to think about 1-indexing. I would rather this be 0-index, and push the complication into error checking.

This is not a strong objection though, because I can just start my relatives at `-1` and treat it like it's 0-index anyways. It just hurts my brain to reason about.

jridgewell (approved):

PR #234, opened 2025-11-13
Range mappings: quick fixes for review feedback

takikawa:
This PR has some relatively minor fixes for #233.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1244,7 +1244,7 @@
           <dl class="header"></dl>
           <emu-alg>
             1. Let _rangeMappings_ be a new empty List.
-            1. If _rawRangeMappings_ is *null*, set _rawRangeMappings_ to *""*.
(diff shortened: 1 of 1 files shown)


takikawa:
Fixed the base commit for this to match the `proposal-range-mappings` branch.

spec.emu
```diff
@@ -1239,12 +1238,13 @@
...
           <h1>
             DecodeRangeMappings (
-              _rawRangeMappings_: a String,
+              _rawRangeMappings_: a String or *null*,
             ): a List of Decoded Range Mapping Offset Records
           </h1>
           <dl class="header"></dl>
           <emu-alg>
             1. Let _rangeMappings_ be a new empty List.
+            1. If _rawRangeMappings_ is *null*, set _rawRangeMappings_ to *""*.
```
jridgewell: You could just return.

PR #235, opened 2025-11-13
Range mappings: refactor decoding out of mapping decoding

takikawa:
This PR addresses some feedback on #233, in particular to move the decoding out of regular mapping decoding.

This new PR's approach has all the range mapping decoding done after regular mapping decoding. Initially all mappings have a `[[IsRangeMapping]]` of false. If there are range mappings encoded, then the next decode step will update the field to true for any range mappings.

If a range mapping is specified for a mapping that doesn't map an original location, or the offset is out of range, then an error is optionally thrown.

One semantics question: should it also optionally throw if the mapping has a [[Name]] too?

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1257,10 +1257,12 @@
           <emu-alg>
             1. For each Decoded Range Mapping Offset Record _rangeMappingOffset_ of _rangeMappingsOffsets_, do
               1. Let _mapping_ be the result of FindMappingForRangeMappingOffset(_mappings_, _rangeMappingOffset_).
-              1. If _mapping_ is ~not-found~, optionally report an error.
-              1. If _mapping_.[[OriginalPosition]] is *null*, optionally report an error.
-              1. Assert: _mapping_.[[IsRangeMapping]] is *false*.
-              1. Set _mapping_.[[IsRangeMapping]] to *true*.
+              1. If _mapping_ is ~not-found~, then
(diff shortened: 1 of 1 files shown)


takikawa:
BTW: this also includes commits from #234 so that should be merged first, then I'll rebase this.

spec.emu
```diff
@@ -1255,21 +1247,44 @@
...
           <emu-alg>
-            1. Let _rangeMappingOffset_ be a new Decoded Range Mapping Offset Record { [[GeneratedLine]] : _generatedLine_, [[MappingIndex]]: _mappingIndex_ }.
-            1. If _rangeMappingOffset_ is an element of _rangeMappings_, then
-              1. Return *true*.
-            1. Else,
-              1. Return *false*.
+            1. For each Decoded Range Mapping Offset Record _rangeMappingOffset_ of _rangeMappingsOffsets_, do
+              1. Let _mapping_ be the result of FindMappingForRangeMappingOffset(_mappings_, _rangeMappingOffset_).
+              1. If _mapping_ is ~not-found~, optionally report an error.
+              1. If _mapping_.[[OriginalPosition]] is *null*, optionally report an error.
```
jridgewell: We need to continue if not-found, else the rest of the steps are an invalid.

jridgewell (approved):

PR #243, opened 2026-03-03
Scopes: Fix 'Decode Scope State Record' initialization

szuend:
We introduced `Index Accumulator Record` and `Position Accumulator Record`. The `Decode Scope State Record` uses these in its definition but we didn't update the `DecodeScopesInfo` abstract operation.


takikawa (approved):
LGTM. I'm reviewing on behalf of Nic after checking in with them.

PR #244, opened 2026-03-03
Scopes: Handle *null* for original variable names

szuend:
This PR turns a broken `"names"` index into an empty string. We need to preserve the length of the array so bindings can be matched. We can't just filter out broken indices.

The alternatives are:
*  Type "Variables" in the original scope as `(string | null)[]` instead of just as `string[]`. This feels wrong though.
* Unconditionally report an error and abort decoding. This feels overly restrictive though.

Drive-by: FIx a typo and use `~none~` consistently.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1271,7 +1271,7 @@
             [[Definition]]: { ... }
             [[Bindings]]: [[
               {
-                [{From]]: { [[Line]]: 0, [[Column]]: 0 },
+                [[From]]: { [[Line]]: 0, [[Column]]: 0 },
                 [[Binding]]: "a",
               },
               {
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -1961,9 +1961,14 @@
             ScopeVariable : Vlq
           </emu-grammar>
           <emu-alg>
-            1. TODO: Handle *null* returned by RelativeName.
-            1. Return « RelativeName(|Vlq|, _accumulator_, _names_) ».
+            1. Let _variable_ be RelativeName(|Vlq|, _accumulator_, _names_).
+            1. If _variable_ is *null*, return « "" ».
+            1. Return _variable_.
```
takikawa: I think this line should be returning `« _variable_ »` to make it a list.

takikawa (approved):
LGTM aside from the one comment

spec.emu
szuend: Good catch! Done.

PR #252, opened 2026-04-08
Range mappings: Change index encoding to start at 0

takikawa:
This implements the change discussed in the April TG4 meeting about reverting the range mapping encoding so that it will start at 0 with relative offset instead of at -1.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -1227,6 +1227,12 @@
             1. Perform DecodeRangeMappingsField of |RangeMappingLineList| with arguments _state_ and _rangeMappings_.
           </emu-alg>
           <emu-grammar>
+            RangeMappingLine : [empty]
+          </emu-grammar>
+          <emu-alg>
+            1. Return.
+          </emu-alg>
+          <emu-grammar>
             RangeMappingList :
               RangeMapping RangeMappingList
           </emu-grammar>
--- a/spec.emu
+++ b/spec.emu
@@ -1246,7 +1246,7 @@
           </emu-grammar>
           <emu-alg>
             1. Let _relativeIndex_ be the VLQUnsignedValue of |Vlq|.
-            1. If _relativeIndex_ = 0 and _state_.[[MappingIndex]] ≠ ~none~, then
+            1. If _relativeIndex_ = 0 and _state_.[[MappingIndex]] is not ~none~, then
               1. Optionally report an error.
             1. If _state_.[[MappingIndex]] is ~none~, then
               1. Set _state_.[[MappingIndex]] to _relativeIndex_.
--- a/spec.emu
+++ b/spec.emu
@@ -1160,7 +1160,29 @@
           </table>
         </emu-table>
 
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -1218,10 +1218,10 @@
           </emu-grammar>
           <emu-alg>
             1. Let _relativeIndex_ be the VLQUnsignedValue of |Vlq|.
-            1. If _relativeIndex_ is 0, then
+            1. If _relativeIndex_ = 0 and _state_.[[MappingIndex]] ≠ 0, then
```
jridgewell: Nit: this means `AA` is valid, both marking the 0th mapping as a range.

jridgewell (approved):

szuend:
One alternative that I forgot about yesterday, is that we could solve this in the grammar itself. Then you don't even need the check. We add a `VlqWithoutZero` or `NonZeroVlq` alongside `Vlq` that doesn't allow `A`.

Then the range mapping grammar is something like:

```
RangeMappingList :
  FirstRangeMapping
  FirstRangeMapping RangeMappingListRest

RangeMappingListRest :
  RangeMapping
  RangeMapping RangeMappingListRest

FirstRangeMapping :
  Vlq

RangeMapping :
  NonZeroVlq
```

Probably a lot of addition to the spec for one use-site. But maybe something to keep in mind if we ever have a need for non-zero VLQs (both signed or unsigned).

takikawa:
I think the `NonZeroVlq` approach is a nice way to do it too, but for now I went with something that adds less to the core grammar.

In the latest push, I initialized the decode state's `[[MappingIndex]]` to `NONE` and then just check that a relative index of `0` can't appear if the state isn't `NONE`. I think this should address @jridgewell's comment about the old check not catching "AA" (thanks for pointing that out!).

spec.emu
```diff
@@ -1160,7 +1160,29 @@
...
+                </th>
+              </tr>
+            </thead>
+            <tr>
+              <td>[[GeneratedLine]]</td>
+              <td>a non-negative integer</td>
+            </tr>
+            <tr>
+              <td>[[MappingIndex]]</td>
+              <td>a non-negative integer or ~none~</td>
```
jridgewell, line 1182: Alternatively we could allow this to be negative, initialize to `-1`, and verify it's positive in appending the new `Range Mapping Offset Record`.

jridgewell (approved):

spec.emu
nicolo-ribaudo: I prefer `~none~`, as it makes the intent clearer when reading.

nicolo-ribaudo (approved):

spec.emu
khan7546jsn-afk: RangeMappingList :
  FirstRangeMapping
  FirstRangeMapping RangeMappingListRest

RangeMappingListRest :
  RangeMapping
  RangeMapping RangeMappingListRest

FirstRangeMapping :
  Vlq

RangeMapping :
  NonZeroVlq

khan7546jsn-afk (commented):
RangeMappingList :
  FirstRangeMapping
  FirstRangeMapping RangeMappingListRest

RangeMappingListRest :
  RangeMapping
  RangeMapping RangeMappingListRest

FirstRangeMapping :
  Vlq

RangeMapping :
  NonZeroVlq

Issue #262, opened 2026-06-06
Scopes: changes to allow more sparse access


rakudrama:
Here is a set of related changes together that can make the stack translation scenario more efficient without impacting other scenarios and without adding weird special cases just for stack translation.

With these changes it should be possible to translate a stack trace without decoding the bulk of the scopes/ranges of a large source map.

## Split _`scopes`_ from _`ranges`_

The tags for scopes are essentially disjoint from the tags for ranges, and the decoding state likewise does not share any state between scopes and ranges. After the split, tag `A` in the scopes and tag `A` in ranges are defined independently. Vendor extensions are shared.

**scopes** is a list of strings parallel to **sources**

`source_map.scopes[i]` holds the encoded original source scopes for the source named in `source_map.sources[i]`.

If the encoded scopes string is empty (`""`) then there are no top level scopes (i.e. **null**). Tag `A` (_EmptyItem_) is no longer needed.  If **scopes** has too few strings, the missing strings are considered to be empty.

**ranges** is a list of strings. Each string encodes a sequence of top level range items. The accumulators start at 'zero' for each string, and a top-level range can not span strings. (A single string is equivalent to a list containing one string).

Top level ranges are monotonic and disjoint in positions covered. For each string in the **ranges** array, the string can be partially decoded to the first _GeneratedRangeStart_ to identify the line and column. A search index of starting positions can be built to identify which of the strings need to be fully decoded to find all the ranges containing a position.

### Scopes grammar changes
_Scopes_ **:**
&emsp; _OriginalScopeTreeList_<sub>opt</sub>

_OriginalScopeTreeList_ :
&emsp; _OriginalScopeTreeItem_
&emsp; _OriginalScopeTreeList_  **,**  _OriginalScopeTreeItem_

Since each file has its own separately encoded scopes, we no longer need to know when we are at the 'next' source file and we can have multiple top-level items. We don't reset any accumulators as we don't cross source files. Vendor extensions can be at top level together with scope trees.

_Scopes_ **:**
&emsp; _OriginalScopeTreeItem_<sub>opt</sub>

_OriginalScopeTreeItem_ :
&emsp; _OriginalScopeTree_
&emsp; _VendorExtensionItem_
&emsp; _InvalidItem_

_OriginalScopeTree_ **:**
[...]

Issue #256, opened 2026-05-23
Scopes: memory requirement for translating stack traces


rakudrama:
There have been discussions about the size of source maps (e.g. https://github.com/tc39/ecma426/issues/41) but mostly concentrated on the file size. This issue is about the memory requirements for using sourcemaps.

Consider translating stack traces to the source language. The application server might do the translation, e.g. when the in-browser app logs a stack trace by sending back to the server. In practice, application servers see unwelcome spikes in memory usage when translating a burst of stack traces. In the server memory profiles I examined, 10-20% of the steady state memory was related to decoded source maps.

One reason for the high memory load  is that the entire source map must be decoded before any part of it can be used. However, translating a stack trace accesses the source map sparsely.

A stack trace typically has [ten or so](https://v8.dev/docs/stack-trace-api#:~:text=holds%20the%20topmost%2010%20stack%20frames) lines. These correspond to a similar number of functions in the original source. The ~10 stack trace lines reference a few generated files (3 is common in the crash logs I sampled). The generated files are large - I looked at a 9.8MB `.js` file that has a 15.8MB source map (12MB if proprietary extensions are stripped).

Digging into this sourcemap:

The **sources** section names ~4.6k files

The **mappings** section is 3.85MB which encodes 530k mapping records.
If the implementation naively represents the decoded source map using classes for each record type in the specification, you might expect one mapping record with its positions to take about 12 words (= 48 bytes when using, say, Java, with compressed pointers).
That would be about 25MB of heap - a 6.6x expansion over the encoded memory footprint.
[Some clever implementations](https://github.com/google/closure-compiler/blob/9f2c9e96d1d11cb74f9439c3f8e2068a4d914eb5/src/com/google/debugging/sourcemap/SourceMapConsumerV3.java#L468) can pack this into arrays and reduce it to ~12MB, or a 3x expansion.

The **names** section has 180k names in 7.9MB. After parsing the JSON, the list and the individual strings it contains occupy about 10MB of heap.

During the stack frame translation we have resident:
* The `.js` file, since it is scanned to find the beginning of the JavaScript function enclosing the stack frame positions (9.8MB)
* The list of names (10MB)
* The decoded mappings (12MB)

That is 31.8MB for only one file and there are typically two more in the stack trace.

### Scopes

[...]

Issue #259, opened 2026-06-03
VLQ inefficiencies and opportunities


rakudrama:
There are two kinds of inefficiency in the VLQ encoding. Multiple VLQ codes decode to the same value, and some decoded values are erroneous in context. This redundancy might be usefully repurposed.

## Duplicate encodings
`A`, `gA`, `ggA`, ... all encode **0**.
`C`, `iA`, `igA`, ... all encode **-1** (signed) or **2** (unsigned).

`B`, `hA`, `ghA`, ... all encode **-2<sup>31</sup>** (signed) or **1** (unsigned). 
(Some decoders get this wrong as a delta of -2<sup>31</sup> is not useful so never appears in practice).

## Erroneous values

The _Mappings_ field uses signed VLQ values for the generated column. Many encodings are wasted, since they would cause GeneratedColumn to become negative.

### Mixed absolute and relative encodings

The Scopes proposal encodes Scope start and end offsets into the original source as unsigned offsets.
Near the end of https://github.com/tc39/ecma426/issues/73#issuecomment-4540562779 is a suggestion how to 'go backwards' to create two scopes covering the same source element. An alternative to this extra element is to enrich VLQ encodings to allow the same position to be encoded. If a VLQ ends with `A` to add a useless high order VLQ digit zero, that is currently redundant. This slightly longer encoding could be used to encode an **A**bsolute value rather than a relative one. The encoding `A` would still be a relative offset of zero.

Sometimes the index into the **names** list needs to go to a very low index. An absolute encoding would be better here than a large negative offset.

### Encode nulls

The signed value -2<sup>31</sup> encoded by `B` will never occur in practice as a signed relative value.
Accessing **names** uses signed relative values. There are some string values that are optional. This could be encoded with `B`.
If there is ever a context where both `null` and -2<sup>31</sup> are valid, the latter could be encoded as `hA`.

### Increased number of relative offsets

A single digit VLQ encodes a signed range of ±15, double digit ±511 etc.
If the signed offset causes the value to become negative, the negative value could be reinterpreted as extending the positive range.
Example: the current position is 0. ±15 has 15 useless values, They could encode +16 to +30.
Example: the current position is 4, the next is 20.  +16 requires a 2-digit VLQ, but if we re-interpret the 11 useless values -15...-5 as +16...+26 then we need only one digit.

[...]

PR #266, opened 2026-06-23
Split "ranges" from "scopes"

rakudrama:
This PR is the first of several to address the following issues

https://github.com/tc39/ecma426/issues/262
https://github.com/tc39/ecma426/issues/256

Generated document:
https://serve-dot-zipline.appspot.com/asset/50212010-f49e-5dd1-9f78-f10a1e1e5b87/zpc/gn6pkhl7ems/

What is included:
* Splitting "scopes" from "ranges"
* "scopes" becomes a list of strings. Each string encodes a list of scopes that correspond with the "source" at the same index.
* By splitting the scopes this way, each "source" can have a list of scopes, avoiding the need for a single top-level scope to contain the other scopes in the file.
* Definitions are now two VLQs, one for the source and one for the scope within that source.

What is not included:
* Relative encoding of call sites
* Splitting encoding of ranges

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -601,7 +601,7 @@
     <li>The <dfn id="json-names"><code>names</code> field</dfn> is an optional list of symbol names which may be used by the mappings field.</li>
     <li>The <dfn id="json-mappings"><code>mappings</code> field</dfn> is a string with the encoded mapping data (see section <emu-xref href="#sec-mappings"></emu-xref>).</li>
     <li>The <dfn id="json-ignoreList"><code>ignoreList</code> field</dfn> is an optional list of indices of files that should be considered third party code, such as framework code or bundler-<emu-not-ref>generated code</emu-not-ref>. This allows developer tools to avoid code that developers likely don't want to see or step through, without requiring developers to configure this beforehand. It refers to the sources field and lists the indices of all the known third-party sources in the source map. Some browsers may also use the deprecated `x_google_ignoreList` field if `ignoreList` is not present.</li>
-    <li><ins>The <dfn id="json-scopes"><code>scopes</code> field</dfn> is an optional list of strings with encoded scope data (see section <emu-xref href="#sec-scopes"></emu-xref>). The encoded scope data strings correspond to source files and are listed in the same order as in the sources field</ins></li>
+    <li><ins>The <dfn id="json-scopes"><code>scopes</code> field</dfn> is an optional list of encoded scope data for each source (see section <emu-xref href="#sec-scopes"></emu-xref>). The contents are listed in the same order as in the sources field. An element of the scopes list is either a string containing encoded scope data, or *null* if there is no scope data for the corresponding source.</ins></li>
     <li><ins>The <dfn id="json-ranges"><code>ranges</code> field</dfn> is an optional string with encoded ranges data (see section <emu-xref href="#sec-scopes"></emu-xref>).</ins></li>
   </ul>
 
@@ -1343,12 +1343,12 @@
           OriginalScopeTree
       </emu-grammar>
 
-      <p>A source with no scopes has its scopes encoded as an empty string.</p>
+      <p>A source with no scopes has its scopes encoded as *null*.</p>
 
       <emu-note example>
         <p>A source map with three authored sources where only the first and third source file have scope information available would be encoded as:</p>
         <pre>
-    "scopes": ["&lt;OriginalScopeTreeList&gt;", "", "&lt;OriginalScopeTreeList&gt;"]
+    "scopes": ["&lt;OriginalScopeTreeList&gt;", null, "&lt;OriginalScopeTreeList&gt;"]
         </pre>
       </emu-note>
 
--- a/spec.emu
+++ b/spec.emu
@@ -667,7 +667,7 @@
         </tr>
         <tr>
           <td><ins>[[FlatScopes]]</ins></td>
-          <td><ins>a List of Original Scope Records or *null*. </ins></td>
+          <td><ins>a List of Original Scope Records or *null*.</ins></td>
         </tr>
       </table>
     </emu-table>
@@ -1634,7 +1634,8 @@
               <td>[[FlatOriginalScopes]]</td>
               <td>a List of Original Scope Records</td>
               <td><ins>*TODO*: Replace with abstract operation on Decoded Source Record?</ins>
-                <p>This is a list of Original Scope Records in order as seen while decoding. This is equivalent to a prefix walk of the scope trees. The |DefinitionScopeIdx| of |RangeDefinition| is used to index into this List.</p></td>
+                <p>This is a list of Original Scope Records in order as seen while decoding. This is equivalent to a prefix walk of the scope trees. The |DefinitionScopeIdx| of |RangeDefinition| is used to index into this List.</p>
+              </td>
             </tr>
             <tr>
               <td>[[ScopePosition]]</td>
@@ -1777,8 +1778,7 @@
           <dt>description</dt>
           <dd>It returns a list of Original Scope Records in order as seen while decoding. This is equivalent to a prefix walk of the scope trees. The |DefinitionScopeIdx| of |RangeDefinition| is used to index into this List.
             <p>*TODO*: It might be cleaner to describe the flattening as an abstract operation. Operations like looking up a scope by source and scope-within-source would operate on the flattened abstraction, which would generally be implemented by caching the flattened list on the Decoded Source Record.</p>
-            <emu-note>There is a chicken-and-egg situation for the generator. Not all scopes in the original program will need to be present in the final source map. The generator will want to emit ranges that encode the scope index into this list, but the index will not be known until the until the surviving (i.e. referenced) scopes are known and the unreferenced scopes removed.</emu-note>
-          </dd>
+            <emu-note>There is a chicken-and-egg situation for the generator. Not all scopes in the original program will need to be present in the final source map. The generator will want to emit ranges that encode the scope index into this list, but the index will not be known until the until the surviving (i.e. referenced) scopes are known and the unreferenced scopes removed.</emu-note></dd>
         </dl>
 
         <emu-alg>
@@ -1800,7 +1800,7 @@
         <h1>
           DecodeGeneratedRanges (
             _ranges_: a String or *null*,
-            _sources_ : a List of Decoded Source Records,
+            _sources_: a List of Decoded Source Records,
             _names_: a List of Strings,
           ): a List of Generated Range Records
         </h1>
@@ -1830,7 +1830,9 @@
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd><p>*FIX*: We don't currently use the result (a list of top-level Original Scope Records). Perhaps we should store the top-level list instead, and specify the lookup in terms of the flattened tree and leave it up to the implementation to determine if some intermediate is cached.</p></dd>
+          <dd>
+            <p>*FIX*: We don't currently use the result (a list of top-level Original Scope Records). Perhaps we should store the top-level list instead, and specify the lookup in terms of the flattened tree and leave it up to the implementation to determine if some intermediate is cached.</p>
+          </dd>
         </dl>
         <emu-grammar>
           Scopes :
@@ -2134,7 +2136,8 @@
           <h1>
             GeneratedRangeCallSite (
               _sources_: a List of Decoded Source Record,
-            ): a Original Position Record</h1>
+            ): a Original Position Record
+          </h1>
           <dl class="header"></dl>
           <emu-grammar>
             GeneratedRangeCallSite :
--- a/spec.emu
+++ b/spec.emu
@@ -714,7 +714,7 @@
         1. <ins>Let _ranges_ be DecodeGeneratedRanges(_rangesField_, _sources_, _namesField_).</ins>
         1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_).
         1. [declared="a,b"] Sort _mappings_ in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if ComparePositions(_a_.[[GeneratedPosition]], _b_.[[GeneratedPosition]]) is ~lesser~.
-        1. <ins>Return the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_, [[Ranges]]: _ranges_.[[Ranges]] }.</ins>
+        1. <ins>Return the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_, [[Ranges]]: _ranges_ }.</ins>
       </emu-alg>
 
       <emu-clause id="sec-GetOptionalString" type="abstract operation">
--- a/spec.emu
+++ b/spec.emu
@@ -587,7 +587,8 @@
     "names": ["src", "maps", "are", "fun"],
     "mappings": "A,AAAB;;ABCDE",
     "ignoreList": [0],
-    "scopes":"BCCAQ,BHAAAC,BCCaC,CDC,CCA,CCY,EDBAA,EGAC,ECvBC,FW,FJ,FI"
+    "scopes": ["BCCAQ,BHAAAC,BCCaC,CDC,CCA,CCY", ""],
+    "ranges": "EDBAA,EGAC,ECvBC,FW,FJ,FI",
   }
   </code></pre>
   </ins>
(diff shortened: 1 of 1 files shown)


szuend:
FYI, no need to self-host. A preview of each PR is created automatically via GitHub actions, the preview for this PR is here: https://tc39.es/ecma426/pr/266/

spec.emu
```diff
@@ -1153,7 +1156,11 @@
...
+    <p>The scopes field is a list of strings, one string per sources entry. The ranges field is a string.</p>
+
+    <p>Each string is a comma (`,`) separated list of items. Each item is a list of base64 VLQs. The first base64 VLQ in each item is a tag that specifies which item is described by the rest of the VLQs. Both original scopes and generated ranges are trees. Each node in these trees are described by a respective "start" and "end" item. "start" items appear in pre-order and "end" items appear in post-order. Any items other then "start" or "end" describe the "current" scope or range. The "current" scope or range is defined by the last seen "start" item.</p>
+
+    <p>*TODO*: Split ranges string into segments to allow partial decoding</p>
```
szuend: nit: Use `<emu-note>` (https://tc39.es/ecmarkup/#emu-note) instead of `<p>` in case we want to leave these in until a follow-up PR (here and below).

spec.emu
```diff
@@ -600,7 +601,8 @@
...
     <li>The <dfn id="json-ignoreList"><code>ignoreList</code> field</dfn> is an optional list of indices of files that should be considered third party code, such as framework code or bundler-<emu-not-ref>generated code</emu-not-ref>. This allows developer tools to avoid code that developers likely don't want to see or step through, without requiring developers to configure this beforehand. It refers to the sources field and lists the indices of all the known third-party sources in the source map. Some browsers may also use the deprecated `x_google_ignoreList` field if `ignoreList` is not present.</li>
-    <li><ins>The <dfn id="json-scopes"><code>scopes</code> field</dfn> is an optional string with encoded scopes data (see section <emu-xref href="#sec-scopes"></emu-xref>)</ins></li>
+    <li><ins>The <dfn id="json-scopes"><code>scopes</code> field</dfn> is an optional list of strings with encoded scope data (see section <emu-xref href="#sec-scopes"></emu-xref>). The encoded scope data strings correspond to source files and are listed in the same order as in the sources field</ins></li>
```
szuend: I'd match `sources` and `sourcesContent` here for consistency and use `null` to signify "no scope info" instead of the empty string.

spec.emu
```diff
@@ -1321,24 +1328,34 @@
...
-      <p>The scopes field is described by the following <em>context-sensitive</em> grammar starting at the |Scopes| goal symbol:</p>
+      <h1>Scopes and Ranges grammar</h1>
+
+      <p>The scopes field is a list of strings, each string corresponding to the the entry from the sources field at the same index. The contents of each string is described by the following <em>context-sensitive</em> grammar starting at the |Scopes| goal symbol:</p>
       <emu-grammar type="definition">
         Scopes :
-          OriginalScopeTreeList
-          TopLevelItemList
-          OriginalScopeTreeList `,` TopLevelItemList
+          OriginalScopeTreeList?
```
szuend, line 1336: This would be a semantic change since we'd now allow multiple top-level "original scopes" per source file, where-as previously we'd restrict it only to one.

This needs to be explicitly called out in the PR description as it goes beyond just splitting scopes/ranges.

spec.emu
```diff
@@ -1747,70 +1767,81 @@
...
+            _scopes_: a String,
             _names_: a List of Strings,
-          ): a Scope Info Record
+          ): a List of Original Scope Records
         </h1>
-        <dl class="header"></dl>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It returns a list of Original Scope Records in order as seen while decoding. This is equivalent to a prefix walk of the scope trees. The |DefinitionScopeIdx| of |RangeDefinition| is used to index into this List.
+            <p>*TODO*: It might be cleaner to describe the flattening as an abstract operation. Operations like looking up a scope by source and scope-within-source would operate on the flattened abstraction, which would generally be implemented by caching the flattened list on the Decoded Source Record.</p>
```
szuend: The spec is not intended to represent an efficient implementation or even guide implementations. It's here to describe the exact semantics as clearly as possible.

In the previous version, it was straight-forward to track a flattened array of scopes to make the indexing clear. But its not a problem at all to just return one (or multiple) scope TREES from `DecodeSourceScopes` and have an abstract operation `GetScopeForScopeIndex` that just does the prefix walk every time to resolve a concrete scopes record for an index. Again, these operations are intended to explain semantics, not prescribe an efficient implementation.

That is also why we don't care about anything like "laziness". That is up to the generator to figure out when and how they want to encode this. We don't have to worry about that here on the spec side.\

spec.emu
```diff
@@ -2079,14 +2111,33 @@
...
-            1. Let _relativeDefinition_ be the VLQSignedValue of |RangeDefinition|.
-            1. Let _scopeIndex_ be AccumulateIndex(_accumulator_, _relativeDefinition_).
-            1. Return _scopes_[_scopeIndex_].
+            1. Return GeneratedRangeDefinition of |RangeDefinition| with arguments _state_ and _sources_.
           </emu-alg>
+          <emu-grammar>
+            RangeDefinition :
+              DefinitionSourceIdx DefinitionScopeIdx
+          </emu-grammar>
+          <emu-alg>
```
szuend, line 2138: nit: I'd have shuffled this around a bit for clarity:

```
            1. Let _sourceOffset_ be the VLQSignedValue of |DefinitionSourceIdx|.
            1. If _sourceOffset_ != 0, set _state_.[[RangeDefinitionScopeIndex]][[Index]] to 0.
            1. Let _scopeOffset_ be the VLQSignedValue of |DefinitionScopeIdx|.
            1. Let _sourceIndex_ be AccumulateIndex(_state_.[[RangeDefinitionSourceIndex]], _sourceOffset_).
            1. Let _scopeIndex_ be AccumulateIndex(_state_.[[RangeDefinitionScopeIndex]], _scopeOffset_).
            1. Let _source_ be _sources_[_sourceIndex_].
            1. Return _source_[[FlatScopes]][_scopeIndex_].
```

szuend (commented):
Initial review comments.

spec.emu
rakudrama: Done.

spec.emu
rakudrama: Is there a useful distinction between "no scope info" and "I have zero scopes"?
I'd rather have one encoding, the shorter one, until we need a distinction.

I don't mind making the change, I just don't think it is 'consistent'. The other two cases have a semantic difference:

For `sources`, the empty string is resolved relative to `sourceRoot`. Reading the root could give you data (like how you get index.html from `http://www.somesite.com`). So `""` is different from `null`. 

For `sourcesContent`, a source could be an empty file. It is possible to encode that position, and it is possible that some language might have code at that position (e.g. there is some initialization code emitted for each source file), so `""` allows the debugger to step there vs `null` making this code non-steppable.

spec.emu
rakudrama: There is a subtle difference - the original code chooses between VLQSignedValue and VLQUnsignedValue.

A different approach compatible with your suggestion is to do something like the section "Increased number of relative offsets" under #259.

rakudrama: It might be better to do something like #259 since it would be more efficient for the signed offset case too, but that would be another PR and another discussion.

spec.emu
```diff
@@ -706,12 +708,13 @@
...
+        1. <ins>Let _scopesField_ be GetOptionalListOfStrings(_json_, *"scopes"*).</ins>
+        1. <ins>Let _rangesField_ be GetOptionalString(_json_, *"ranges"*).</ins>
+        1. <ins>Let _sources_ be DecodeSourceMapSources(_baseURL_, _sourceRootField_, _sourcesField_, _sourcesContentField_, _ignoreListField_, _namesField_, _scopesField_).</ins>
+        1. <ins>Let _ranges_ be DecodeGeneratedRanges(_rangesField_, _sources_, _namesField_).</ins>
         1. Let _mappings_ be DecodeMappings(_mappingsField_, _namesField_, _sources_).
         1. [declared="a,b"] Sort _mappings_ in ascending order, with a Decoded Mapping Record _a_ being less than a Decoded Mapping Record _b_ if ComparePositions(_a_.[[GeneratedPosition]], _b_.[[GeneratedPosition]]) is ~lesser~.
-        1. <ins>Return the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_, [[Ranges]]: _scopesAndRanges_.[[Ranges]] }.</ins>
+        1. <ins>Return the Decoded Source Map Record { [[File]]: _fileField_, [[Sources]]: _sources_, [[Mappings]]: _mappings_, [[Ranges]]: _ranges_.[[Ranges]] }.</ins>
```
hbenl: `[[Ranges]]` should be set to `_ranges_` instead of `_ranges_.[[Ranges]]`

spec.emu
szuend: Wouldn't the `sourcesContent` argument then not also hold for "scopes"? `""` is "empty file/empty scopes", vs `null` marking it as "no scopes information available / non-steppable".

IMHO the 2 extra characters are not really a consideration. It compresses well either way for transport.

spec.emu
szuend: Ah, I missed the signed/unsigned difference depending on the "sourceOffset". Thats very subtle but might be wroth doing!

rakudrama:
Are more changes going to be requested or is this PR ready to land?

spec.emu
```diff
@@ -1747,70 +1760,74 @@
...
+            _scopes_: a String,
             _names_: a List of Strings,
-          ): a Scope Info Record
+          ): a List of Original Scope Records
         </h1>
-        <dl class="header"></dl>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It returns a list of the top level Original Scope Records. Nested scopes are reachable via the children.
+            <emu-note>There is a chicken-and-egg situation for the generator. Not all scopes in the original program will need to be present in the final source map. The generator will want to emit ranges that encode the scope index into this list, but the index will not be known until the until the surviving (i.e. referenced) scopes are known and the unreferenced scopes removed.</emu-note></dd>
```
szuend: Superfluous `the until`

spec.emu
```diff
@@ -2263,21 +2332,24 @@
...
+          1. <ins>If _index_ &lt; _encodedScopes_' length, then</ins>
+            1. <ins>Let _decodedSourceScopes_ be DecodeSourceScopes(_encodedScopes_[_index_], _names_).</ins>
+            1. <ins>Set _decodedSource_.[[RootScopes]] to _decodedSourceScopes_.</ins>
           1. Append _decodedSource_ to _decodedSources_.
         1. Return _decodedSources_.
       </emu-alg>
 
       <emu-note>Implementations that support showing source contents but do not support showing multiple sources with the same URL and different content will arbitrarily choose one of the various contents corresponding to the given URL.</emu-note>
+
+      <emu-note>Implementations that support showing source contents but do not support showing multiple sources with the same URL and different content will arbitrarily choose one of the various contents corresponding to the given URL.</emu-note>
```
szuend, line 2354: Duplicated `<emu-note>`?

szuend (commented):
The PR overall lgtm, but there are 2 points we still need to decide, and I'd like @jridgewell  and @hbenl to weigh in:

1. Whether to use `null` or the empty string `""` in the `scopes` string array of the source map JSON to signify "no scope information" present. As mentioned in the comments, I'm leaning more towards consistency with `sources`/`sourcesContent` and use `null`.

2. The "definition reference" (how a range points to a scope) is now two VLQs. The source index and inside that source file, the scope index. The PR proposes to interpret the scope index VLQ as either a signed or an unsigned VLQ depending on whether the source index is zero or non-zero respectively. This is the first instance in the spec where we use different "signed-ness" of a VLQ depending on some circumstance.

If we decide to keep it, we should add a `<emu-note>` calling it out, as its a bit subtle.

szuend:
Please note that once this PR will land, we'll also have to update the respective conformance tests here: https://github.com/tc39/source-map-tests/tree/main/decoding/scopes. We can wait until all the other PRs you have in the pipeline (the `ranges` splitting) also have landed.

hbenl:
In reply to https://github.com/tc39/ecma426/pull/266#pullrequestreview-4690826723:
1. I would also prefer `null` over `""`
2. I think this makes sense and would like to keep it

rakudrama:
> The PR overall lgtm, but there are 2 points we still need to decide, and I'd like @jridgewell and @hbenl to weigh in:
> 
> 1. Whether to use `null` or the empty string `""` in the `scopes` string array of the source map JSON to signify "no scope information" present. As mentioned in the comments, I'm leaning more towards consistency with `sources`/`sourcesContent` and use `null`.

I have changed it to `null`.

> 2. The "definition reference" (how a range points to a scope) is now two VLQs. The source index and inside that source file, the scope index. The PR proposes to interpret the scope index VLQ as either a signed or an unsigned VLQ depending on whether the source index is zero or non-zero respectively. This is the first instance in the spec where we use different "signed-ness" of a VLQ depending on some circumstance.
> 
> If we decide to keep it, we should add a `<emu-note>` calling it out, as its a bit subtle.

I have added the note.

spec.emu
rakudrama: fixed

spec.emu
```diff
@@ -1321,24 +1328,34 @@
...
           OriginalScopeTreeList `,` OriginalScopeTreeItem
 
         OriginalScopeTreeItem :
           OriginalScopeTree
-          EmptyItem
+      </emu-grammar>
 
-        EmptyItem :
-          `A`
+      <p>A source with no scopes has its scopes encoded as *null*.</p>
```
jridgewell:
```suggestion
      <p>A source with no scopes has its scopes entry encoded as *null*.</p>
```

spec.emu
```diff
@@ -105,7 +105,7 @@
     <li>Support server-side stack trace deobfuscation</li>
   </ul>
   <p>The <emu-not-ref>original source</emu-not-ref> map format (v1) was created by Joseph Schorr for use by Closure Inspector to enable source-level debugging of optimized JavaScript code (although the format itself is language agnostic). However, as the size of the projects using source maps expanded, the verbosity of the format started to become a problem. The v2 format (Source Map Revision 2 Proposal) was created by trading some simplicity and flexibility to reduce the overall size of the source map. Even with the changes made with the v2 version of the format, the source map file size was limiting its usefulness. The v3 format is based on suggestions made by Pavel Podivilov (Google).</p>
-  <p>The source map format does not have version numbers anymore, and it is instead hard-coded to always be "3".</p>
+  <p>The source map format does not have version numbers anymore, and it is instead hard-coded to always be “3”.</p>
```
jridgewell, line 108: There's a lot of stray quotation mark changes.

spec.emu
jridgewell: Still there?

spec.emu
```diff
@@ -1498,6 +1506,9 @@
           GeneratedRangeCallSite :
             `I` CallSiteSourceIdx CallSiteLine CallSiteColumn
 
+          RangeDefinition :
+            DefinitionSourceIdx DefinitionScopeIdx
```
jridgewell, line 1510: Should we separate `DefinitionSourceIdx` into its own top level instruction? That way if we have 10 ranges that are within the same source, they don't each need to encode the source index. We just emit a ``` `J` DefinitionSourceIdx ``` whenever we want to change the current source.

jridgewell (approved):
Sorry for the delayed review.

> Whether to use null or the empty string "" in the scopes string array of the source map JSON to signify "no scope information" present. As mentioned in the comments, I'm leaning more towards consistency with sources/sourcesContent and use null.

I'd lean to `null` like the others.

> The "definition reference" (how a range points to a scope) is now two VLQs. The source index and inside that source file, the scope index. The PR proposes to interpret the scope index VLQ as either a signed or an unsigned VLQ depending on whether the source index is zero or non-zero respectively.

Seems fine to me.

spec.emu
rakudrama: The quotes come from emu-format, and without them it fails the PR checks.

spec.emu
rakudrama: The `RangeDefinition` is part of `GeneratedRangeStart`, and is optional and indicated by one of the `RangeFlags`.
The `J` item would have to occur before the `GeneratedRangeStart`.

The `DefinitionSourceIdx` is relative so the same-source encoding is always `A`.
Your suggestion wins when we skip enough `A`s to to exceed the size of `JX,` preceding the `GeneratedRangeStart`, i.e. four in a row. If the source changes every range, it costs 2 bytes more per range for all the extra `J?,` items.

It would be very much dependent of the code. For the Dart compilers, there is lots of inlining, so he region definitions bounce around. For a simpler 1-1 transpiler, not so much.

It is really not clear, can we revisit this later?

szuend: I'd also suggest we revisit this later.

szuend (approved):

spec.emu
```diff
@@ -1747,70 +1760,74 @@
         </emu-clause>
       </emu-clause>
 
-      <emu-clause id="sec-DecodeScopesInfo" type="abstract operation">
+      <emu-clause id="sec-DecodeSourceScopes" type="abstract operation">
         <h1>
-          DecodeScopesInfo (
-            _scopes_: a String or *null*,
+          DecodeSourceScopes (
+            _scopes_: a String,
```
hbenl: `_scopes_` may be `null`

spec.emu
```diff
@@ -1747,70 +1760,74 @@
...
+      <emu-clause id="sec-DecodeGeneratedRanges" type="abstract operation">
+        <h1>
+          DecodeGeneratedRanges (
+            _ranges_: a String or *null*,
+            _sources_: a List of Decoded Source Records,
+            _names_: a List of Strings,
+          ): a List of Generated Range Records
+        </h1>
+        <dl class="header"></dl>
+        <p>*TODO*: move this to just before DecodedGeneratedRangeTrees.</p>
```
hbenl, line 1800: any reason not to do that in this PR?

hbenl (approved):

